PR #241, opened 2012-06-08
tests/test-tls-session-resume.lua: Added test for TLS session resumption

mkandrashoff:
Didn't have that test before.


zhaozg:
It's ancient, so close.

PR #258, opened 2012-06-24
Rough draft of how the docs may look.

robsearles:
Updated lib/luvit/querystring.lua to include inline docs in GitHub
Markdown format. There are a couple of points about this discussed
below. Also included is a stand-alone document for the querystring
module, again in GitHub Markdown format.

Regarding the inline docs, I have chosen to demark them using a
triple dash block comment, i.e: ---[[ .... ---]]. This is so it is
easier to include code snippets in the docs, which themselves may
include block comments using the default double dash format.

This does work to some effect, however the current doc scraper misses
out the first doc block comment and I'm not sure why, also some stray
double dashes appear.

### DIFF
--- a/docs/querystring.md
+++ b/docs/querystring.md
@@ -1,0 +1,77 @@
+# Query String
+
+Helper for parsing, encoding and decoding query strings
+
+## querystring.parse(str, [sep], [eq])
+
+Parses a querystring, returning table of url decoded tokens.
+
+Optionally override the default separator ('&') and assignment ('=') characters.
+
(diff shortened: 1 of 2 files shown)


kaustavha:
So the latest in the Docs series of developments is this repository:
https://github.com/luvit/luvit-docs
Its derived from nodejs' docs and works similarly. 
They currently provide a decent high level overview of luvit, albeit much remains to be done.

zhaozg:
It's ancient, so close.

Issue #267, opened 2012-06-30
http request doesn't emit end from HTTP/1.0 server


tim-smart:
Doing some last.fm api stuff for a small tool that gets the currently playing track. Currently running into an issue where the end event is not being emitted.

See https://gist.github.com/3021609 (working from the querystring.stringify branch) for a reproducible test case.

creationix:
I think this is fixed with the new http-codec library used in luvit 2.0.  Please re-test.

tim-smart:
Retested and ran into an issue where the response event isn't being emitted: #664

tim-smart:
Moved the callback to the second argument of `http.request`, seems to work great but the event loop sits on idle after everything finishes. I expected it to exit the program!

creationix:
@tim-smart maybe it's a keepalive connection?  We didn't clone the entire agent behavior from node's http API.  Can you share your example?

tim-smart:
Here is a gist:
https://gist.github.com/9727132763b264051fa8

creationix:
I see.  I wonder what node does.  Perhaps we can uv.unref the connection while it's idle and then ref it again if it's used again.  We already close all active handles on process exit.

rphillips:
hmm ```uv.walk(p)``` does not show an active tcp handle

tim-smart:
Created a new issue, will close this one as the `end` event is now working fine! Thanks

PR #268, opened 2012-06-30
HTTP refactor

tim-smart:
This aims to bring a more feature rich http client and server without sacrificing too much performance. It makes the api very similar to node internally and externally.

Currently the performance is almost as fast as before the patch, but I haven't done extensive benchmarking. We are currently a lot slower than node 0.8 at small responses, but get faster after about 100kb. We will need to continue to comb over the code and optimize where we can.

See also #267


rphillips:
Thank you, Tim, for the hard work on this PR. The HTTP subsystem has been rewritten in the 2.0 release.

PR #307, opened 2012-08-16
tests:test-fs-mkdir, native API change:fsOpen takes mode as int

hnakamur:


tests/test-fs-mkdir.lua
```diff
@@ -0,0 +1,80 @@
...
+    assert(err == null)
+    assert(FS.existsSync(pathname) == true)
+    ncalls = ncalls + 1
+  end)
+
+  process:on('exit', function()
+    unlink(pathname)
+    assert(ncalls == 1)
+  end)
+end)();
```
creationix: Did you know you can just use `do .. end` to create local scope in lua?  No need for self-calling functions like in JavaScript.

creationix:
I just noticed this PR is against a the old verison of luvit.  I'm closing for now unless the old version still need this added.

PR #493, opened 2014-08-28
WIP: Bump tools for Windows Compatibility

rjemanuele:
* Requires luajit at tag v2.0.3
* Requires gyp at latest?


rjemanuele:
replaced with cmake in Luvit 2

PR #510, opened 2014-10-01
WIP: Pathing fixes/improvements (mostly for Windows)

squeek502:
Work in progress. Any feedback is appreciated.

Note that the issue fixed by https://github.com/squeek502/luvit/commit/01c8a864da62608c78b2d71a7ffbca42cf013be8 was making the test runner unable to spawn child processes correctly, resulting in the inability to run any tests at all (all tests would have the exit code -1 and no stdout/stderr data).

Note also that I'm (mostly) trying to conform to the [node.js path test cases](https://github.com/joyent/node/blob/master/test/simple/test-path.js). I'll make sure to detail any differences with how node.js handles things when the PR is potentially ready to be merged.


rjemanuele:
i think this is ready, +1

rphillips:
+1

squeek502:
Note that the things listed in a previous comment are still issues for Windows paths, namely:
* UNC paths do not currently enforce having both a hostname and sharename to be valid
* Windows paths that start with a separator are not currently treated as absolute, but they should be (or, rather, semi-absolute; they should be treated as relative to the cwd's root)

rjemanuele:
@squeek502 Neither of those missing features differ from the original version though, correct?

squeek502:
Correct.

PR #577, opened 2015-01-03
Fix http chunked encoding

gsick:
Hi all,

Fix in http.lua for http methods other than get, head and connect when useChunkedEncodingByDefault is true.
The size calculation was wrong.


creationix:
Thanks, I guess I've been doing too much JS over the years.

gsick:
Have you planned a minor release soon?

creationix:
I think @rphillips wanted to cut a release soon.

rphillips:
yeah, we need to tag master

gsick:
Ok, nice. Thanks

PR #578, opened 2015-01-04
Fix: Use-after-free and subsequent memory leaks

mihe:
Freeing the FS request data seems to [free the request itself](https://github.com/luvit/luvit/blob/9bd578462c034b24b60af42b148061c5fad5c48b/src/luv_fs.c#L194-199), which messes with [the cleanup](https://github.com/joyent/libuv/blob/6e24ce23b1e7576059f85a608eca13b766458a01/src/win/fs.c#L1961) in libuv. This should fix it and allow it to deallocate the path properly.


mihe:
I should probably have mentioned that this only seems to be an issue when launching from (for example) the Visual Studio debugger, which marks freed memory with 0xfeeefeee.

rphillips:
thanks!

PR #591, opened 2015-01-26
Fix require on windows

grandao:
Drive letter on windows can be lower case too


rphillips:
thank you!

grandao:
there is one more fix comming

PR #596, opened 2015-01-29
add Error object to stream_writable

rphillips:


creationix:
Should probably replace all the `core.Error` instances to use the new alias.  Property lookups are more expensive in luajit than V8. (different optimizing techniques)

creationix:
+1

PR #601, opened 2015-02-02
fix(tls): add release_buffers mode

rphillips:


app/modules/_common_tls.lua
```diff
@@ -254,7 +255,7 @@ function TLSSocket:_read(n)
         local i, o = self.out:pending()
         if i > 0 then
           net.Socket._write(self, self.out:read(), nil, function()
-            handshake()
+            timer.setImmediate(handshake)
```
creationix: Is the `_write` callback sync sometimes?  Or is there some other reason to need to defer handshake?

rphillips: _write shouldn't be sync, I can revert this change.

rphillips: fixed

creationix:
+1

PR #605, opened 2015-02-06
RFR: Start to move all modules to lit

creationix:


james2doyle:
It looks like [this window require fix was undone](https://github.com/grandao/luvit/commit/110401e9fe48d270872d0501ad32ed3a3316adf8) in the `app/modules/require.lua` file

creationix:
@james2doyle good eye.  I installed the version from the lit version.  I guess the change in luvit never got published there.  I'll keep rebasing this PR till it's all working.  Then it will be obvious which packages are out of data on lit.

creationix:
All modules in `app/modules` are now tagged and published to lit.luvit.io.  To test it, you can `rm -rf modules && lit install` in the app folder.  It should delete all the files and replace them with exact copies.

creationix:
The next proposed step is to add inter-dependencies metatadata for all these modules and move them all to their own independent git repositories.  Then the luvit repo will quite empty except for package.lua, main.lua and init.lua.

PR #607, opened 2015-02-09
Enable string formatting operator

imzyxwvu:
This adds a string formatting operator that works like the Python one without affecting behaviors of most already-existing Lua code.

The following expressions will work the same:

("%d + %d = %d"):format(1, 2, 3)
"%d + %d = %d" % { 1, 2, 3 }


rphillips:
wdyt of: https://github.com/luvit/luvit/pull/608

Issue #610, opened 2015-02-12
Port Signal Support to Luvit2


rphillips:

creationix:
Move to the misc.c file and misc section of luv.c and then +1

PR #611, opened 2015-02-12
feat(process): kill and getpid

rphillips:


creationix:
+1 if it passes

rphillips:
Fixed #610

PR #612, opened 2015-02-13
RFR: HTTP/HTTPS client

rjemanuele:
HTTP/HTTPS client support
* url parser
* associated tests

### DIFF
--- a/Make.bat
+++ b/Make.bat
@@ -5,14 +5,17 @@ IF NOT "x%1" == "x" GOTO :%1
 :luvit
 IF NOT EXIST lit.exe CALL Make.bat lit
 ECHO "Building luvit"
-lit.exe make app
+SET LUVI_APP=app/
+SET LUVI_TARGET=luvit.exe
+lit\luvi-binaries\Windows\luvi.exe
+SET "LUVI_APP="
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+*~
```
creationix: Is this for temp files created by your editor?  If so, I think it's better to have this in your global system gitignore.

```sh
git config --global core.excludesfile ~/.gitignore_global
```

I only mention because I use a variety of editors and if I added gitignore entries on every project I worked on for the cruft they all leave, it would be quite significant.  Sorry if I assume incorrectly.

rphillips: nice tip... never knew

rphillips:
a few lint errors
```
    app/modules/http.lua:215:9: unused variable i
    app/modules/http.lua:215:12: unused variable header
    app/modules/http.lua:216:7: shadowing definition of variable i on line 215
    app/modules/http.lua:216:10: shadowing definition of variable header on line 215
    app/modules/http.lua:217:16: unused variable value
    app/modules/http.lua:239:9: unused variable req
```

rphillips:
otherwise... +1

PR #614, opened 2015-02-16
fix(test): timeout test

rphillips:
* Set to super low timeout
* Fix the expect calls


app/modules/net.lua
```diff
@@ -74,10 +74,6 @@ function Socket:bind(ip, port)
   uv.tcp_bind(self._handle, ip, tonumber(port))
 end
 
-function Socket:_onTimeoutReal()
```
rphillips: <dead code> : not used anymore

creationix:
+1

PR #618, opened 2015-02-17
add fs.readstream functionality

rphillips:


creationix:
Don't assume offset will be used.  Some fd types only work with implicit offsets.  This is done by passing nil to luvit's `fs.read` function.  Change the default to be `offset = nil` and only increment or calculate using offset if it's set.  Some options like length may require offset to default to 0 for their calculations to work.  Ideally you could track bytes consumed without actually sending offsets to libuv in that case.

rphillips:
Fixed

creationix:
Looks better, but I think if you pass in `length`, but not `offset`, it will ignore the length.  You might need to count bytes read in this case and use that instead of offset to track progress.

rphillips:
thanks tim... +1

PR #624, opened 2015-02-19
change set_verify to verify_mode

rphillips:


creationix:
Use `curl -L https://github.com/luvit/lit/raw/0.9.3/web-install.sh | sh` to install lit and it should give you the right version.

PR #625, opened 2015-02-19
Library tidy

creationix:
This first part flattens the tree a bit by merging app with the root and adding the needed filters on the files.


rphillips:
+1

PR #629, opened 2015-02-20
http codec doesn't treat "" as nil in encodeHead.

DBarney:
when replying to an http request like so `res:finish("")` it should have the same behavior as `res:finish()`. Right now it just crashes trying to get a status code from "".

This causes "" to be treated like nil.

### DIFF
--- a/modules/http-codec.lua
+++ b/modules/http-codec.lua
@@ -88,7 +88,7 @@ exports.encoder = function ()
   local encodeHead, encodeRaw, encodeChunked
 
   function encodeHead(item)
-    if not item then return end
+    if not item or item == "" then return end
     local head, chunkedEncoding
     local version = item.version or 1.1
     if item.method then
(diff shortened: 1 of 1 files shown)


creationix:
Why should `""` be treated the same as `nil`?  They have very different meanings in http-codec.  In the decoder, `nil` means the socket is disconnected while `""` means the end of a request or response body (but there may be more data thanks to HTTP keepalive)

modules/http-codec.lua
```diff
@@ -88,7 +88,7 @@ exports.encoder = function ()
   local encodeHead, encodeRaw, encodeChunked
 
   function encodeHead(item)
-    if not item then return end
+    if not item or item == "" then return end
```
creationix: I think the correct thing here is to `return item`.  If the user passes in `""` to the encoder it's basically a no-op and shouldn't end the stream like `nil` does.  By emitting an empty string we're essentially telling the socket to write nothing.

creationix: Also we should probably check the type of item and throw an error with a helpful error message if it's not a table (After this line where `nil` and `""` are valid inputs)

DBarney:
fair enough. I'll make the change.

PR #630, opened 2015-02-20
socket errors on a http request need to be re-emitted by the http request object.

DBarney:
No too sure if there is a different approach I should take.  But this does cause errors like ECONNREFUSED to be emitted on a http request object.

Also the test is returning an error, but it isn't ECONNREFUSED like i expected it to be. Which is why it is just checking against nil right now.


rphillips:
lgtm, thanks!

PR #631, opened 2015-02-23
Add binary extension to search paths

creationix:


rphillips:
@creationix thoughts on the binary_module/ search path? With my setup, the `modules/sigar.*` got always ignored.

creationix:
On unix binary shared libraries are stored in `lib` folders.  What about using `libs` or `lib` for binary modules.

rphillips:
works for me... i'll fix that up.

creationix:
Actually, I kinda like the idea of having two different folders for searching.  One can be `lib` for libraries statically included (not provided by lit) and the other can be `opt` for where lit puts things.  Lit will exclude `opt` when importing modules, but not `lib`.

What do you think?

creationix:
`opt` contains locally installed software. Originated in System V, which has a package manager that installs software to this directory.

creationix:
Implemented in https://github.com/luvit/lit/commit/6acd8db2c84ac36f7c99bd46ff68ba53b2a2e9bd.  This PR is obsolete.

PR #633, opened 2015-02-24
Add more to the HTTP/HTTPS tests

rjemanuele:
* test the stream end


modules/http.lua
```diff
@@ -222,9 +223,21 @@ function ClientRequest:initialize(options, callback)
...
     end
+    local uafound = key:lower() == 'user-agent'
+    if uafound then
+      user_agent_found = value
+    end
     table.insert(self, header)
   end
 
+  if not user_agent then
+    user_agent = 'luvit/http/' .. exports.version .. ' luvi/' .. luvi.version
```
creationix: Should calculate the default value only once outside the closure.  It's not going to change for the lifetime of the process.

creationix:
Rebase on luvi-up since http-lax has been merged?

rjemanuele:
rebased

modules/http.lua
```diff
@@ -207,11 +208,18 @@ end
 local ClientRequest = Writable:extend()
 exports.ClientRequest = ClientRequest
 
+function exports.ClientRequest.getDefaultUserAgent()
+  if exports.ClientRequest.default_user_agent ~= nil then
+    exports.ClientRequest.default_user_agent = 'luvit/http/' .. exports.version .. ' luvi/' .. luvi.version
```
rphillips: needs a space in-between the two

rphillips: should be == nil

modules/http.lua
```diff
@@ -351,17 +371,20 @@ function ClientRequest:done(data, encoding, cb)
   end
 end
 
-function exports.request(options, onResponse)
+function exports.parse_url(options)
```
rphillips: parse_url should probably be `parse` from the nodejs docs.

modules/http.lua
```diff
@@ -222,9 +230,21 @@ function ClientRequest:initialize(options, callback)
...
+      user_agent_found = value
+    end
     table.insert(self, header)
   end
 
+  if not user_agent then
+    user_agent = self.getDefaultUserAgent()
+  end
+
+  if user_agent ~= '' then
```
rphillips: let's always put the user-agent header

modules/http.lua
rjemanuele: it is url.parse like node.  this is more of an internal helper for http to simplify string detection

modules/http.lua
rjemanuele: it does always put it unless there is one in the headers provided by the user already

creationix:
+1

PR #638, opened 2015-02-25
done no longer ignores data being passed before the connections is established

DBarney:
calling done before the connection was established causes the request to be sent without any data. This will save the data until the connection is established and then send it then.


rjemanuele:
nice find.  +1

PR #644, opened 2015-02-28
fixes(http stream): re-emit finish event

rphillips:


rphillips:
according to node this object should emit a close event

rjemanuele:
+1 nice fix

PR #653, opened 2015-03-09
expose package version in -v and --version

james2doyle:
All this PR does is expose the `package.lua` version in the -v/--version flag. It follows the same convention as the `lit version` function.

```sh
$ luvit -v
luvit version: 1.9.5
$ lit version
lit version: 0.11.4
```

I figured reading the package would be nicer than hard-coding, since it's one less thing to bump when the version increments.


rphillips:
+1. thanks

PR #666, opened 2015-03-12
fix(http): flush the stream with an _end()

### LINKED ISSUE #665: When `http.request` finishes, the process doesn't exit
Here is a reproducible case: https://gist.github.com/9727132763b264051fa8

From #267

rphillips:
* Fixes #665


creationix:
Didn't we want to rename all the `_end` methods with `finish`?

rphillips:
yes, we need to do that

PR #669, opened 2015-03-12
Fix HTTP chunked request termination, test chunked http requests

rjemanuele:


deps/http.lua
```diff
@@ -237,6 +239,9 @@ function ClientRequest:initialize(options, callback)
     if uafound then
       user_agent = value
     end
+    if key:lower() == 'transfer-encoding' then
```
creationix: This is now unused, should we remove it?

rphillips:
+1

PR #674, opened 2015-03-16
Luvit1: add no compression and release buffers to ssl context

rphillips:
Reference: https://journal.paul.querna.org/articles/2011/04/05/openssl-memory-use/


rjemanuele:
+10 at least

PR #683, opened 2015-03-24
fix handle nil error

gsick:
Hi all,

Just a small fix for this error message:
 [string "bundle:/init.lua"]:52: attempt to index local 'handle' (a nil value)

Thanks

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -48,7 +48,7 @@ return function (main, ...)
 
   -- When the loop exits, close all unclosed uv handles.
   uv.walk(function (handle)
-    if not handle:is_closing() then handle:close() end
+    if handle and not handle:is_closing() then handle:close() end
   end)
   uv.run()
 


rphillips:
i wonder why uv.walk would return a nil... thanks for the patch

creationix:
@gsick how can you reproduce this?  walk is never supposed to return nil.

rphillips:
Perhaps there is a handle in the event loop, and it gets freed by lua's gc (unref'ed)

gsick:
@creationix,

Sorry, I didn't give any details. I looked deeper and I was lucky to have this nil case instead of seg fault.
I have libuv handles created and attached to the loop in C. But, in luvit 2, it seems that luv expect to have handles created only in lua. In the callback luv_walk_cb, a luv_handle_t object is expected, maybe a test on active/close/closing with the handle flags (with uv bitmask uv__handle_active, uv_closed...) could be add here or somewhere else if you don't want duplicate the existing test in luvit init.lua. Native libuv uv.walk function doesn't return nil but closed or not active handles.

How to reproduce: just create a timer in c, add it to the loop, close it. Run with luvit. [gist](https://gist.github.com/gsick/bc8cf50db811898fa0d4)

It would be nice to keep the possibility to share the same evt loop between lua and C luvit library, but should it be manage in luv ? without too much complexity?
I'm trying to migrate [here](https://github.com/gsick/crazy-snail) a C module for luvit 1 to luvit 2. It's the only issue I found. I would like to not rewrite everything, let me know if you know better way to handle this case or if you know how to fix it in luv.

PR #690, opened 2015-04-09
Added the Emitter:propagate method to allow event propagation similar to pipe

GabrielNicolasAvellaneda:
Just a simple but useful change. I've also added an example.

### DIFF
--- a/deps/core.lua
+++ b/deps/core.lua
@@ -302,6 +302,13 @@ function Emitter:wrap(name)
   end
 end
 
+-- Propagate the event to another emitter.
+function Emitter:propagate(eventName, target)
+  if (target and target.emit) then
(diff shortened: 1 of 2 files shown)


rphillips:
nice. I like the API.

GabrielNicolasAvellaneda:
Hello Ryan, I would like to be a maintainer on Luvit, Can It be possible? 
I'm working on a project that is based on Luvit and I'm detecting some bugs and I want to provide some fixes. Also I've some ideas to add to it. 

Thanks!
Gabriel.-

creationix:
@GabrielNicolasAvellaneda come join us in #luvit on freenode irc where we discuss things (if your timezone permits, we're mostly in central and pacific US timezones).  I'm very open to new contributors, we just need to make sure we have a common vision for project direction before I can grant commit access.

PR #691, opened 2015-04-11
Added a test suite for the url module. Also updated url.parse to be more like nodejs. Allowing parsing more urls formats. Also added a test show that http.get does not execute 'end' event.

GabrielNicolasAvellaneda:


creationix:
Great Stuff!

Could you squash and rebase this?  Also it seems to be causing more tests to fail, is this intended?

GabrielNicolasAvellaneda:
Tim, I will rebase the changes.
That test that is failing is a proof that there is a problem with the http implementation. This is a bug that I detected. (test-http-stream-finish.lua) I can comment this test for now if you want but I think that will be nice to have this test. I will also work this week on solving the bug on the http.
What do you think?

creationix:
Yes, new failing tests for existing bugs are fine.  I just wanted to make sure there weren't any actual regressions.

deps/url.lua
```diff
@@ -25,22 +25,28 @@ function exports.parse(url, parseQueryString)
   local chunk, protocol = url:match("^(([a-z0-9+]+)://)")
   url = url:sub((chunk and #chunk or 0) + 1)
 
-  local chunk, auth = url:match('(([0-9a-zA-Z]+:?[0-9a-zA-Z]+)@)') 
+  chunk, auth = url:match('(([0-9a-zA-Z]+:?[0-9a-zA-Z]+)@)')
```
creationix: `auth` here is a leaking global.  I highly recommend running luacheck on all lua code to prevent scope leakage.  I'm using sublimelinter with luacheck in my workflow and it's super helpful.

GabrielNicolasAvellaneda: Thanks! Good to know about your workflow and tools. I was used vim for my lua coding. I will give a try to sublimelinter and luacheck.

creationix: I believe @rphillips used vim for a while and probably knows how to add luacheck to it.

GabrielNicolasAvellaneda: Thanks for the info!. I will ask him.

rphillips: You will want the [Syntastic](https://github.com/scrooloose/syntastic) module.
```
let g:syntastic_lua_checkers = ["luac", "luacheck"]
```

GabrielNicolasAvellaneda: Perfect! Thanks Ryan

rphillips:
@GabrielNicolasAvellaneda Can you rebase this once more?

GabrielNicolasAvellaneda:
Yes no problem.

GabrielNicolasAvellaneda:
Done rebasing. Now there are 2 commits, do you want that I rebase the two commits into one?

GabrielNicolasAvellaneda:
I will cleanup my repo first.

PR #693, opened 2015-04-13
fix(childprocess): close handle on a process kill

rphillips:


creationix:
+1

What does `:destroy()` do anyway?

rphillips:
destroy closes the pipes... we might want to refactor that.

PR #694, opened 2015-04-13
fix(childprocess): propagate errors through the pipes

rphillips:


creationix:
+1 when the tests are done

PR #698, opened 2015-04-20
fix(net): double callback in net :_write

### LINKED ISSUE #697: Socket:_write() executes the callback twice in case of error
Causes uncaught exception:

```
Uncaught exception:
[string "../stream/stream_wr..."]:349: attempt to call local 'cb' (a nil value)
stack traceback:
	[string "../stream/stream_wr..."]:349: in function 'onwriteError'
	[string "../stream/stream_wr..."]:371: in function 'onwrite'
	[string "../stream/stream_wr..."]:137: in function 'callback'
	[string "../net.lua"]:97: in function <[string "../net.lua"]:94>
```

net.lua
```lua
function Socket:_write(data, encoding, callback)
  if not self._handle then return end
  timer.active(self)
  uv.write(self._handle, data, function(err)
    timer.active(self)
    if err then
      callback(err) -- <- second execution when we got an error
      self:destroy(err)
      return
    end
  end)
  callback() -- <- always executed
end
```

rphillips:
fixes #697


creationix:
+1

tests/test-http-stream-finish.lua
```diff
@@ -5,10 +5,6 @@ require('tap')(function(test)
     local server
     server = http.createServer(function (req, res)
       local body = "Hello world\n"
-      res:on("finish", expect(function()
```
rjemanuele: Removing this, removes the test for "finish"

tests/test-http-post-1mb.lua
```diff
@@ -42,9 +42,9 @@ require('tap')(function(test)
       request:on('end', expect(function()
         p('server:onEnd')
         assert(postBuffer == data)
-        response:write(data)
```
rjemanuele: Moving server:close() seems ok but we shouldn't need to wrap the response:finish() in the callback.

deps/http.lua
```diff
@@ -107,9 +108,9 @@ function ServerResponse:flushHeaders()
   self:writeHead(self.statusCode, self.headers)
 end
 
-function ServerResponse:write(chunk)
```
rjemanuele: nice

deps/http.lua
```diff
@@ -122,8 +123,8 @@ function ServerResponse:finish(chunk)
   if #last > 0 then
     self.socket:write(last)
   end
-  self.socket:shutdown(function()
```
rjemanuele: what does the nextTick here fix?

tests/test-http-stream-finish.lua
rphillips: fixed. put this back in.

creationix:
+1

rjemanuele:
+1

PR #702, opened 2015-04-21
Added functionality to the url.parse method and the related test suite

GabrielNicolasAvellaneda:
Now you can merge :+1:

### DIFF
--- a/tests/test-http-stream-finish.lua
+++ b/tests/test-http-stream-finish.lua
@@ -1,7 +1,8 @@
 local http = require("http")
 
 require('tap')(function(test)
-  test('http stream end', function(expect)
+
+  test('http stream end #1', function(expect)
     local server
     server = http.createServer(function (req, res)
       local body = "Hello world\n"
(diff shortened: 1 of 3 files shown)


tests/test-http-stream-finish.lua
```diff
@@ -1,7 +1,16 @@
 local http = require("http")
 
 require('tap')(function(test)
-  test('http stream end', function(expect)
+
```
rphillips: commented out test?

GabrielNicolasAvellaneda: Yes, I've commented because is a test for a bug. I will remove the comment with the fix of the http bug. What do you think?

rphillips: removing it seems fine... and renaming 'http stream end #2' back to #1.

GabrielNicolasAvellaneda: I can do this, but i will add this test again later. I will do what if you
want, just confirm this again :)

2015-04-20 23:03 GMT-03:00 Ryan Phillips <notifications@github.com>:

GabrielNicolasAvellaneda: @rphillips I have just removed it and renamed http strean end #2 back to #1

PR #705, opened 2015-04-28
Make https.request fall back to http.request for non-https protocols

squeek502:
* Allows for using https without needing to parse the url beforehand to choose between http.request and https.request

Changes something like:
```lua
local url = require('url')
local https = require("https")
local http = require("http")

local test_url = "http://luvit.io"
local parsed_url = url.parse(test_url)
local prot = parsed_url.protocol == "https" and https or http

prot.get(parsed_url, function() end)
```
to:
```lua
local https = require("https")

local test_url = "http://luvit.io"

https.get(test_url, function() end)
```

Also just noticed https://github.com/luvit/luvit/pull/704, which this will conflict with. Will rebase this once that is merged.


creationix:
This is an issue with the node api which we're mirroring here in luvit.  I suppose there is no harm in making the `https` and `http` modules both auto-detect the protocol based on the string, but it just makes the api weird then.

rphillips:
This also would require someone using the https module to validate they are connected over `tls` with an attribute on the socket, which seems like a vulnerability.

squeek502:
Fair enough. Would it make sense to add
```
  if options.protocol and options.protocol == 'https' then
    error(fmt('Protocol %s not supported', options.protocol))
  end
```
to http.request, then?

The reason for this pull request is that I found it confusing trying to figure out which request method I should use, and http.request failing on https:// urls does not provide much useful feedback.

For example:
```lua
local http = require("http")

local req = http.request('https://api.github.com/users/squeek502/events/public', function (res)
  p(res.statusCode, res.headers)
  res:on('data', function (chunk)
    p("ondata", {chunk=chunk})
  end)
  res:on('end', function ()
    p('stream ended')
  end)
end)
req:done()
```
outputs nothing except
```
ECONNREFUSED: connection refused
```

while the same code with `"https://luvit.io"` as the url outputs:
```
301     { server = 'nginx/1.6.2', ['content-type'] = 'text/html', connection = 'close', ['content-length'] = '184', location = 'https://luvit.io/', date = 'Tue, 28 Apr 2015 17:32:59 GMT' }
'ondata'        { chunk = '<html>\r\n<head><title>301 Moved Permanently</title></head>\r\n<body bgcolor="white">\r\n<center><h1>301 Moved Permanently</h1></center>\r\n<hr><center>nginx/1.6.2</center>\r\n</body>\r\n</html>\r\n' }
'stream ended'
```

rphillips:
hmm. we do have this module ```https://github.com/virgo-agent-toolkit/luvit-request/blob/master/lib/request.lua``` and it's in lit. We do need some error reporting though.

creationix:
When I hit this issue in node I put a conditional around which module to use.

You could do something like:

```lua
local req = (url:match("^https") and https or http).request(url, ...)
```

My suggestion is to not use the node style API directly, it's poorly designed and the only reason luvit copies it is to make migration easy for people coming from node.

squeek502:
Alright. Will an abstraction like https://github.com/virgo-agent-toolkit/luvit-request/blob/master/lib/request.lua be added to luvit? Seems like it'd be a good thing.

rphillips:
Probably not... Tim's approach is a oneliner and fits the use case.

squeek502:
Fair enough. :smile:

PR #709, opened 2015-04-29
readline: Fix case where multiple control characters get entered at once

squeek502:
For example: 
```lua
'\027[D\027[D\027[D\027[D\027[D\027[D\027[D'
```

This PR makes readLine consume a key at a time instead of doing simple string comparisons against the whole input buffer

Note: There is still a lingering broken edge case when extremely large inputs are sent. If they are so large that the onKey call is broken up into multiple calls, then the cutoff point is arbitrary, meaning that it can end at something like '\027' and then the next call would start with '[D', which would then cause the trailing \027 to go unhandled and then the proceeding [D would be treated as normal characters and inserted into the Editor's line.

I was testing this using a strange method, as it just so happens that ConEmu sends massive amounts of arrow key presses if you click somewhere in the window that is far away from the cursor. It seems like it should be possible to add a test case for this, though, so let me know if you'd like me to do that.

EDIT: Went ahead and added test cases. Note that the last 4 asserts in the `readline Editor:onKey` test would have failed before this PR.

### DIFF
--- a/deps/readline.lua
+++ b/deps/readline.lua
@@ -269,9 +269,19 @@ local function escapeKeysForDisplay(keys)
   end)
 end
 
-function Editor:onKey(key)
-  local char = string.byte(key, 1)
-  if     char == 13 then  -- Enter
+-- an array of tables so that the iteration order is consistent
+-- each entry is an array with two entries: a table and a function
+-- the table can contain any number of the following:
+--   numbers (to be compared to the char value),
+--   strings (to be compared to the input string that has been truncated to the same length),
+--   functions (to be called with the (key, char) values and returns either the consumed keys or nil)
+-- the function recieves the Editor instance as the first parameter and the consumedKeys as the second
+--   its returns will be propagated to Editor:onKey if either of them are non-nil
+--   note: the function is only called if the key handler is the one doing the consuming
+local keyHandlers = 
+{
+  -- Enter
+  {{13}, function(self)
     local history = self.history
     local line = self.line
     -- Only record new history if it's non-empty and new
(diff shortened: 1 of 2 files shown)


creationix:
Sounds like we need to use a decoder to decouple the input chunks from the key press events.  That would clean up the abstraction here and fix the edge case where a key is split up.

If you haven’t seen the "decoder" style, it's essentially a function that takes input and returns the first event as well as any leftover bytes.  If there isn't enough data for even one event, it returns nothing.

For example, here is a decoder that splits up newline delimited input into line events.

```lua
local function deline(input)
  local p = input:find("\n", 1, true)
  -- If there is no newline, wait for more data
  if not p then return end
  return input:sub(1, p), input:sub(p + 1)
end
```

And to consume this in an evented system, you could do:

```lua
local buffer = ""
local function onChunk(chunk)
  buffer = buffer .. chunk
  repeat
    local event, extra = deline(buffer)
    if event then
      emit(event)
      buffer = extra
    end
  until not event
end
```

creationix:
Oops, I forgot the best part of this style.  Once you have a decoder and an evented decorator, you can use it to drop-in replace an emitter handler:

```lua
stream:on("data", lines(function (line)
end))
```

And instead of getting events for every chunk, you get events for every line, already parsed and reassembled.  We should be able to do the same thing for terminal control commands.

squeek502:
Sounds good, but that's going to require a more robust control sequence parser to be able to reliably tell what is an unfinished control sequence waiting on more input and what isn't.

For reference: 
* http://en.wikipedia.org/wiki/ANSI_escape_code#Sequence_elements
* http://en.wikipedia.org/wiki/C0_and_C1_control_codes

For example, the F1 key sends '\027[[A' and is totally unhandled at the moment, so that would need to be treated as a delimeter rather than an unfinished control sequence. Otherwise, there'd be no way to differentiate it from something like a '\027[' that happens to be an unfinished '\027[D'.

But, I think it's enough of an edge case (and will take quite a bit more effort to solve) that it should be a separate pull request; I'll add an issue for it so it doesn't get lost. 

For now, I'll rebase this to solve the conflict with #709.

squeek502:
Split the overflow edge case to #714. This should be ready to be pulled.

PR #710, opened 2015-04-29
readline: Add Editor:insertAbove to allow output without disrupting the prompt

squeek502:
* Also made the unhandled key print use insertAbove


creationix:
Very cool.

PR #716, opened 2015-04-29
fix tipo: executible -> executable

lduboeuf:


creationix:
Thanks, I saw that the other day and forgot to fix it.

PR #720, opened 2015-05-01
add fs.appendFile

rphillips:


zhaozg:
test-fs-append-file.lua and test-fs-append-file-sync.lua, please

deps/fs.lua
```diff
@@ -577,3 +577,31 @@ end
...
+  callback = callback or function() end
+  local function write(fd, offset, buffer, callback)
+    local function onWrite(err, written)
+      if err then return fs.close(fd, function() callback(err) end) end
+      if written == #buffer then
+        fs.close(fd, callback)
+      else
+        if offset == -1 then offset = 0 end
+        offset = offset + written
+        write(fd, offset, buffer, callback)
```
creationix: Do we still want to write the full buffer in this case or do we want a substring of it?

rphillips: nice catch

deps/fs.lua
```diff
@@ -577,3 +577,30 @@ end
 function fs.createReadStream(path, options)
   return fs.ReadStream:new(path, options)
 end
+function fs.appendFile(filename, data, callback)
+  callback = callback or function() end
+  local function write(fd, offset, buffer, callback)
+    local function onWrite(err, written)
+      if err then return fs.close(fd, function() callback(err) end) end
+      if written == #buffer then
```
creationix: This is still going to be a problem with partial writes since `#buffer` is always the full buffer, but you only write part in the second write.

deps/fs.lua
```diff
@@ -577,3 +577,30 @@ end
...
+        fs.close(fd, callback)
+      else
+        offset = offset + written
```
creationix: maybe do `buffer = buffer:sub(offset)` so that the `#buffer` above matches what we're trying to write?

rphillips: fixed in a12fcbb

creationix:
+1

PR #727, opened 2015-05-05
Fix extension match

Playermet:
Extension match fails on Windows, if location path contains folder with dots.


creationix:
Thanks!

PR #732, opened 2015-05-11
Connection upgrade support

sousoux:
Hi. I added connection upgrade support in http and as a result found a few other little things in ReadableStream and Socket. The only one that I'm not too sure about is the socket resume call to _read(0) which seems absolutely necessary to get things started again but is clearly not correct. Would love a different way of getting the same effect.

### DIFF
--- a/deps/http.lua
+++ b/deps/http.lua
@@ -70,13 +70,14 @@ local ServerResponse = Writable:extend()
 exports.ServerResponse = ServerResponse
 
 function ServerResponse:initialize(socket)
+  Writable.initialize(self)
   local encode = codec.encoder()
   self.socket = socket
   self.encode = encode
   self.statusCode = 200
   self.headersSent = false
   self.headers = {}
-  for _, evt in pairs({'close', 'finish'}) do
+  for _, evt in pairs({'close', 'finish', 'drain', 'end' }) do
     self.socket:on(evt, utils.bind(self.emit, self, evt))
   end
 end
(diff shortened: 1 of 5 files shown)


creationix:
Thanks for working on this.  I never had the motivation to port node's websocket API since I feel it was flawed in design.

Just out of curiosity, have you seen how I handle websockets in my coroutine based libraries?  Weblit (a web framework) now has websocket support https://github.com/creationix/weblit.  Weblit is structured somewhat like express.js in that different layers of "middlewares" handle requests and responses in turn.  The websocket handler is part of this system instead of a separate `upgrade` event like in node.  This means that http auth, cookies, logging, etc all apply to websocket requests the same as all other http requests.

sousoux:
Hi Tim

I did look at the coroutine based libraries.

I am already using the utopia framework on Luvit 0.83 so I have moved all
of that across (along with modifications that I made before). The websocket
stuff I've done fits into that as a middleware so I get cookies/logging/etc.

Since I have quite a way to go before I get everything across to 2.X I
didn't want to change too much at this stage.

Martin

creationix:
@rphillips what do you think about this one?

rphillips:
I think this looks good... As long as travis passes. +1

creationix:
@sousoux I think you need to rebase your PR.  Sorry for being slow slow in reviewing it.

sousoux:
Done. With a little trouble!

PR #738, opened 2015-06-04
fix(tls): export socket and copy _connecting

rphillips:


creationix:
Why bump the build versions of so many packages?

rphillips:
going to release a new version

rphillips:
lit add libs/* said the versions were out of date

PR #743, opened 2015-06-12
fix bus in tls module, variable scope error

zhaozg:


rphillips:
thanks!

deps/tls/common.lua
```diff
@@ -245,8 +245,9 @@ function TLSSocket:_read(n)
       return self:destroy(err)
     elseif cipherText then
       if self.inp:write(cipherText) then
+        local plainText
         repeat
-          local plainText = self.ssl:read()
+          plainText, extra = self.ssl:read()
```
creationix: I don't understand this fix.  The new `extra` variable isn't declared anywhere and so we're just setting to a global.  Also the `repeat` .. `until` block in lua can see variables from inside the loop in it's conditional.  There is no need to move the local to the outside.

deps/http-codec.lua
```diff
@@ -264,18 +263,15 @@ exports.decoder = function ()
...
     -- If the entire chunk fits, pass it all through
     if length <= bytesLeft then
       bytesLeft = bytesLeft - length
       return chunk, ""
     end
 
-    return sub(chunk, 1, bytesLeft), sub(chunk, bytesLeft + 1)
+    length = bytesLeft
+    bytesLeft = 0
+    return sub(chunk, 1, length), sub(chunk, length + 1) or ''
```
creationix: I don't think the `or ''` is needed here.  `String.sub` always returns a string which is always truthy.

deps/tls/common.lua
zhaozg: Sorry for my knowledge partly. :)

zhaozg:
According @creationix advice, I think luvit team should change the code.

creationix:
@zhaozg No need to feel shy.  My background is just in scripting languages and not C, each person has unique strengths.

I'm just trying to understand the change.  What's the bug this is fixing?

I can see the internal behaviour changes slightly in http-codec.lua, but as far as I can tell, it's the same external behaviour.

I never bothered to set `bytesLeft` to `0` in all cases because it's *always* reset before entering the state.

zhaozg:
Yes, nothing fixed, this is a joke. please @rphillips help me to do some fix this wrong commit.
```shell
git reset --hard 6d71e1318c890433246600495c912876c83d7afe
git push --force
```

creationix:
@zhaozg Thanks for clarifying.  And don't feel bad, we appreciate all your contributions.

It's hard-reverted now.

zhaozg:
Luvit is great project and include nice coders.

PR #745, opened 2015-06-12
fix(tls): fix resume of TLS socket

rphillips:


deps/tls/common.lua
```diff
@@ -222,18 +222,12 @@ function TLSSocket:sni(hosts)
...
-  end
-  i = self.out:pending()
-  if i > 0 then
-    net.Socket._write(self, self.out:read(), callback)
+  if not self.ssl then return end
+  self.ssl:write(data)
+  while self.out:pending() > 0 do
+    net.Socket._write(self, self.out:read())
   end
+  callback()
```
creationix: This is a sync callback now.  Probably now the behaviour we want.

PR #747, opened 2015-06-13
fix(http): flush res with a nil push

rphillips:


zhaozg:
will fire event on('data',chunk==nil),  Don't fire event data equals nil when socket close in my test case.

rphillips:
@zhaozg can you link me the test case?

zhaozg:
Please wait my test case,I'm busy now

zhaozg:
@rphillips 
I think should fire data nil, then fire _socketEnd.

```lua
local timer = require('timer')
local net = require('net')
local uv = require('uv')

local function createTestServer(port, host, listenCallback, hanleCallback)
  local server = net.createServer(function(client) hanleCallback(client) end)
  server:listen(port, host, listenCallback)
  server:on("error", function(err) assert(err) end)
  return server
end

require('tap')(function (test)
  test("nil Data", function(expect)
    local port = 10081
    local host = '127.0.0.1'
    local server
    server = createTestServer(port, host, expect(function()
      local client
      client = net.createConnection(port, host, expect(function()
        client:on('data', expect(function(data)
          assert(data==nil)
          print('shoud fire')
          client:destroy()
          server:close()
        end))
        client:on('error',function(err) 
          print('on("error") fired')
        end)
        client:on('_socketEnd',function(err) 
          print('on("_socketEnd") fired')
        end)
        
      end))
    end),
    function(accpeted) 
      accpeted:destroy()
    end)
  end)
end)
```

rphillips:
luv will emit a nil on the data event, but luvit will emit an 'end' event instead.

```lua
local timer = require('timer')
local net = require('net')
local uv = require('uv')

local function createTestServer(port, host, listenCallback, hanleCallback)
  local server = net.createServer(function(client) hanleCallback(client) end)
  server:listen(port, host, listenCallback)
  server:on("error", function(err) assert(err) end)
  return server
end

local port = 10081
local host = '127.0.0.1'
local server
server = createTestServer(port, host, function()
  local client
  client = net.createConnection(port, host, function()
    client:on('data', function(data)
      print('shoud fire')
    end)
    client:on('error',function(err) 
      print('on("error") fired')
    end)
    client:on('_socketEnd',function(err) 
      print('on("_socketEnd") fired')
    end)
    client:on('end',function(err) 
      print('on("end") fired')
      client:destroy()
      server:close()
    end)

  end)
end,
function(accpeted) 
  accpeted:destroy()
end)

```

zhaozg:
Got it thanks

PR #749, opened 2015-06-13
The `fileCache cache the loaded data, not state.

AndrewTsao:


creationix:
@rphillips this will fix a bug where module:load will sometimes return `true` instead of the actual data, but it also means that all required files will now have their contents cached in memory.  Will this cause a problem for you?

I could just change module:load to not use this cache.

@AndrewTsao I'm saving the state and not data on purpose to save memory, but as mentioned above, I've noticed a bug.  Is this the same bug you saw or are you changing this for a different reason?

AndrewTsao:
I find this bug when I read luvit how to load module.  I think this cache is redundant, it is never hit. because of there is a `ModuleCache(https://github.com/luvit/luvit/blob/master/deps/require.lua#L258) on the top of this layer.

creationix:
@AndrewTsao right, because of the moduleCache, the require system will never try to load the fileCache data.  I just need to fix module:load to not use this cache.

PR #751, opened 2015-06-17
A few little fixes

sousoux:
1) There is no way to disable colors at present. This patch helps one do that.

2) Possibility of hang since callback is not called if socket is destroyed

3) Currenly no was to require('bundle:/absolute/path/to/something')

This one requires a look. Should it be bundle://absolute?

### DIFF
--- a/deps/net.lua
+++ b/deps/net.lua
@@ -137,7 +137,7 @@ end
 
 function Socket:shutdown(callback)
   if self.destroyed == true then
-    return
+    return callback()
   end
 
   if uv.is_closing(self._handle) then
--- a/deps/require.lua
+++ b/deps/require.lua
@@ -229,9 +229,10 @@ end
 
 function Module:resolve(name)
   assert(name, "Missing name to resolve")
-  if name:byte(1) == 46 then -- Starts with "."
+  local debundled_name = name:match("^bundle:(.*)") or name
+  if debundled_name:byte(1) == 46 then -- Starts with "."
     return fixedRequire(pathJoin(self.dir, name))
-  elseif name:byte(1) == 47 then -- Starts with "/"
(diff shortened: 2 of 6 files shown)


zhaozg:
Like point 2

creationix:
> 1) There is no way to disable colors at present. This patch helps one do that.

Sure there is, just set it to `false`.  Try running `luvit -n` on the command line to see it in action.

sousoux:
True. That is dumb. The other two are valid though I think.

deps/net.lua
```diff
@@ -137,7 +137,7 @@ end
 
 function Socket:shutdown(callback)
   if self.destroyed == true then
-    return
+    return callback()
```
creationix: Good catch.

deps/require.lua
```diff
@@ -229,9 +229,10 @@ end
 
 function Module:resolve(name)
   assert(name, "Missing name to resolve")
-  if name:byte(1) == 46 then -- Starts with "."
+  local debundled_name = name:match("^bundle:(.*)") or name
+  if debundled_name:byte(1) == 46 then -- Starts with "."
     return fixedRequire(pathJoin(self.dir, name))
-  elseif name:byte(1) == 47 then -- Starts with "/"
+  elseif debundled_name:byte(1) == 47 then -- Starts with "/"
```
creationix: Nice fix.  I was wondering why this didn't work.

creationix:
> That is dumb.

Well, I needed a value that means to use whatever color scheme is default for the platform.  I have logic elsewhere to try and detect if the console supports 256 colors or just 16 colors.

There was a bug before where it was impossible to disable colors, but it was fixed by checking for exactly `nil` instead of just any falsy value.

sousoux:
I meant I was being dumb ... not you!
Thanks for merging the other two

PR #753, opened 2015-06-24
process: flush stdout on exit

rphillips:


creationix:
So you emit exit first, but don't actually exit till after the flush?

rphillips:
i think that should give anything in the exit handler time to run... what do you think?

creationix:
It does give the exit handler an undetermined time to run.  I fear this will cause more race conditions than anything.

rphillips:
fixed

creationix:
+1

creationix:
Should we do the same thing for stderr?

PR #754, opened 2015-06-24
stdout/stderr flush on exit

rphillips:


deps/process.lua
```diff
@@ -108,13 +108,31 @@ local function removeListener(self, _type, listener)
...
+
+  local function onCompleted()
+    if completed == 2 then
+      self:emit('exit', code)
+      os.exit(code)
+    end
+  end
+
+  local function onStdoutEnd()
+    completed = completed + 1
```
creationix: You should move the increment into the shared `onCompleted()` function since it's shared among all paths.

creationix: In fact, you can then just have a single shared `onEnd` callback that you directly use for both the streams.

creationix:
```lua
local left = 2
local function onEnd()
  left = left - 1
  if left > 0 then return end
  self:emit('exit', code)
  os.exit(code)
end

process.stdout:once('end', onEnd)
process.stderr:once('end', onEnd)
```

rphillips: good call. tweaked

creationix:
+1

PR #756, opened 2015-06-25
avoid crash if nil string in http request:write()

gsick:
Hi,

It's just for add more safety in case of nil string/body.
Original exception:
```
Uncaught exception:
[string "bundle:deps/http.lua"]:407: attempt to get length of local 'encoded' (a nil value)
```
Thanks


rphillips:
thanks. +1

PR #748, opened 2015-06-13
tls support client session reuse

zhaozg:
This RP make tls support client session resuse, and need update lua-openssl to newest master branch of https://github.com/zhaozg/lua-openssl.

Please review this PR should be accept?
to get better performance TLSSocket should use shared secureContext, eg: 
secureContext = tls.createCredentials(...) --to share will multi TLSSocket,
tls.connect({secureContext = secureContext, ...},onconnect_cb)

PR #758, opened 2015-06-25
tls support client session reuse

rphillips:
Same PR as https://github.com/luvit/luvit/pull/748


zhaozg:
luvi not include lua-openssl support tls client session resuse yet.

PR #759, opened 2015-06-26
stream doesn't send close event

gsick:
Hi,

In process:exit() the callback onClose is never call because stream doesn't send "close" event but "end" event.
I didn't change the version...

Thanks

### DIFF
--- a/deps/process.lua
+++ b/deps/process.lua
@@ -116,9 +116,9 @@ local function exit(self, code)
     self:emit('exit', code)
     os.exit(code)
   end
-  process.stdout:once('close', onClose)
+  process.stdout:once('end', onClose)
   process.stdout:_end()
-  process.stderr:once('close', onClose)
+  process.stderr:once('end', onClose)
   process.stderr:_end()
 end
 


rphillips:
in this case, the stdout/stderr streams should emit a close event because the underlying socket is a tty

```lua
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process:exit(0)
```

rphillips:
Make sure you are using 2.2.2 as well.

gsick:
I'm using the master branch. I cleaned everything and checkout master.

I added a simple trace in the callback onClose:
```lua
local function exit(self, code)
  ...
  local function onClose()
    print("onClose")
   ...
    self:emit('exit', code)
    os.exit(code)
  end
  ...
end
```

After I execute your example:
```lua
process.stdout:write('hello world\n')
process.stdout:write('hello world\n')
process:exit(0)
```

the output is:
```bash
hello world
hello world
```

the onClose cb is not executed

rphillips:
@gsick could you try this branch out for me `fixes/fix_stdout_stderr_wait_for_finish` and let me know?

gsick:
@rphillips Tested and it works like a charm.

PR #762, opened 2015-07-02
shutdown fixes

rphillips:
Martin, can you try this branch out?

\cc @sousoux

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -35,6 +35,16 @@ return function (main, ...)
     if jit.os ~= 'Windows' then dns.loadResolver() end
   end
 
+  -- EPIPE ignore
+  do
+    local jit = require('jit')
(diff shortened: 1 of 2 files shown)


rphillips:
Hmm. Really close

rphillips:
cool. that worked.

init.lua
```diff
@@ -35,6 +35,16 @@ return function (main, ...)
     if jit.os ~= 'Windows' then dns.loadResolver() end
   end
 
+  -- EPIPE ignore
+  do
+    local jit = require('jit')
+    if jit.os ~= 'Windows' then
+      local sig = uv.new_signal()
+      uv.signal_start(sig, 'sigpipe', function() end)
```
creationix: No need for an empty function.  It also accepts nil which translates to SIGIGN.

PR #763, opened 2015-07-03
require now works with ffi libs.

DBarney:
not sure if this is the best approach, but it does work.

### DIFF
--- a/deps/require.lua
+++ b/deps/require.lua
@@ -30,7 +30,8 @@ local luvi = require('luvi')
 local bundle = luvi.bundle
 local pathJoin = luvi.path.join
 local env = require('env')
-local os = require('ffi').os
+local ffi = require('ffi')
+local os = ffi.os
 local uv = require('uv')
 
 local realRequire = _G.require
@@ -183,6 +184,17 @@ local function moduleRequire(base, name)
 end
 
 
+local function load_lib(path, fnName)
+  fn, err = package.loadlib(path, fnName)
+  if not fn then
+    fn, err = ffi.load(path)
+  else
+    fn = fn()
+  end
+  return fn, err
+end
+
+
 local moduleCache = {}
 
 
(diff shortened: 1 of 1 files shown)


creationix:
What's the goal here?  It basically just overloads require to also support `ffi.load` which is pretty easy to use already.  I guess it's because you want to use require's relative logic and search paths to load normal DLL's?

Also I have one issue with the implementation.  The `err` from the loadlib is always obscured by the error from ffi.load.  I'd rather the first error be returned in the case both attempts fail.

DBarney:
I have been bundling and loading DLL files from the bundle. Yes you are right it is very easy to use `ffi.load`, the problem is all the step that have to happen before then:

1 figure out where in the bundle the DLL is stored
2 read it
3 create a temporary file on the disk (where ever that is)
4 write the data to the disk
5 finally load the file using `ffi.load`

Those steps are already in the `require()` logic, I figured a better approach then replicating the code in an alternate require specifically `for ffi.load`, would be to extend this require to support loading and returning DLLs that were already loaded with `ffi.load`.

I was also a little torn about which error to return. I will change it to the first one.

creationix:
Hmm, so before we had the `error(realPath .. "#" .. fnName .. ": " .. err)` error message that really helped debug issues with native addons.  I'd like to see that someone preserved.  I can see how it doesn't apply when trying to use ffi.load.

creationix:
BTW, do you know about `module:action(path, callback)`?  That is a generic mechanism that does most of what you need I think.

creationix:
See https://github.com/creationix/lit-gamepad/blob/master/init.lua#L12-L14

DBarney:
I did not know about that, that actually works much better then changing how require works. Thanks for the link!

All I do now is this `local lib = module:action(path, ffi.load)` and it works automatically, which is exactly what I was looking for with the require change.

I will close the commit the module:action function allows is much more flexible.

creationix:
I'm glad it worked.  Sorry for the general lack of documentation.

PR #765, opened 2015-07-06
Add support TLS client session reuse

zhaozg:
This will make luvit support TLS client session reuse, please see tests/test-tls-connect-session-reuse.lua

Server side session reuse already support, this can be verify by 

```bash
openssl s_client -connect host:port -reconnect -no_ticket | grep Session-ID
```
session-id should keep same,

### DIFF
--- a/deps/tls/common.lua
+++ b/deps/tls/common.lua
@@ -145,8 +145,14 @@ function TLSSocket:_init()
   self.inp = openssl.bio.mem(8192)
   self.out = openssl.bio.mem(8192)
   self.ssl = self.ctx.context:ssl(self.inp, self.out, self.server)
-  if (not self.server) and self.options.servername then
-    self.ssl:set('hostname',self.options.servername)
+
+  if (not self.server) then
+    if self.options.servername then
+      self.ssl:set('hostname',self.options.servername)
+    end
+    if self.ctx.session then
+      self.ssl:session(self.ctx.session)
+    end
   end
 end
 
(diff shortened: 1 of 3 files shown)


zhaozg:
appveyor failed,because not use luvi not support session_reused check
```
 
# Starting Test: https-client - https-client
Uncaught Error: [string "C:/projects/luvit/deps/tls/common.lua"]:164: attempt to call method 'session_reused' (a nil value)
Packaging artifacts...OK
Uploading artifact lit.exe (2.7 MB)...OK
Uploading artifact luvit.exe (2.8 MB)...OK
```

zhaozg:
travis failed becuase with same reason
```

# Starting Test: https-client - https-client
Uncaught Error: [string "/home/travis/build/luvit/luvit/deps/tls/commo..."]:164: attempt to call method 'session_reused' (a nil value)
make: *** [test] Error 255
```

zhaozg:
please see at https://github.com/zhaozg/lua-openssl/blob/master/src/ssl.c#L1720-L1726
and https://github.com/zhaozg/lua-openssl/blob/master/src/ssl.c#L1840

rphillips:
Thanks for the note... We do need to release a new luvi with this support. Sorry that it's taken a few false starts to get this right.

rphillips:
Luvi 2.1.4 is built and should work with this now.

rphillips:
the new test passes for me on my OSX machine.

PR #768, opened 2015-07-08
tweaks for error objects in fs module

rphillips:


creationix:
+1

rjemanuele:
+1

deps/utils.lua
```diff
@@ -83,6 +85,7 @@ local function adapt(c, fn, ...)
         assert(coroutine.resume(c, ...))
       end
     else
+      if e then e = Error:new(e) end
       err, data = e, {...}
```
creationix: These two lines could be simplified to:

```lua
err, data = e and Error:new(e), {...}
```

PR #770, opened 2015-07-09
change tmp directory to /var/run

rphillips:


creationix:
+1

creationix:
So is `/var/run` a better tmp dir?  I thought root was required to write to it.

PR #776, opened 2015-07-17
Add headerMeta to make http headers smarter (fixes #773)

### LINKED ISSUE #773: HTTP client is too sensitive
The http client seems to only accept headers as nested tables (as opposed to keyed maps).  Also it doesn't switch to chunked encoding when there is no Content-Length included.

creationix:
This pulls in the magic metatable from weblit that makes working with headers much easier.

It's even better than the node API now.  You can do case-sensitive lookups and replacements of headers using simple key/value access.

```lua
-- Will add a new header header containing {"Content-Length", "100"}
-- It will also remove any pre-existing headers with the same name (case insensitive)
headers["Content-Length"] = 100

-- Here we can do a case-insensitive lookup of headers without having to lowercase our key.
if headers["Content-Length"] then
  -- do something
end
```

The internal storage is always a plain array with `{key, value}` entries.  The metatable handles everything else.

You can even replace several headers (of the same name) with several new headers (of the same name) at once by passing a table of strings as the value.

```lua
headers["Set-Cookie"] = {
  "one cookie",
  "another cookie",
  "more cookies"}
```

### DIFF
--- a/deps/http.lua
+++ b/deps/http.lua
@@ -38,16 +38,57 @@ local date = require('os').date
 local luvi = require('luvi')
 local utils = require('utils')
 
+
+-- Provide a nice case insensitive interface to headers.
+-- Pulled from https://github.com/creationix/weblit/blob/master/libs/weblit-app.lua
+local headerMeta = {
+  __index = function (list, name)
+    if type(name) ~= "string" then
+      return rawget(list, name)
+    end
+    name = name:lower()
+    for i = 1, #list do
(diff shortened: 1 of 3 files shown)


rphillips:
+1

creationix:
Also this PR changes header behavior to not actually send the headers till the first `write()` or `finish()`.  This way we can always know when writing out the head if there is a body or not.  Before the logic sometimes got confused and didn't auto-set chunked encoding.

PR #777, opened 2015-07-21
Binary certs

creationix:
This adds a make target `make update-certs` that downloads the mozilla root certs, parses them with the go utility and converts the format to a compact DER database instead of lua code.

The tls module is then updated to use this compact format.

This change shrinks the code size significantly for the luvit library and might reduce memory overhead too since the certs never pass through the vm as base64 lua strings or source code.


rphillips:
very nice, +1

PR #778, opened 2015-07-22
Add support for 'size 0' files where the kernel lies

creationix:
Sometimes the stat call says a file has size of zero, but will return data if you read anyway.  This patch attempts to read a file in streaming mode whenever the stat is zero.

This is in line with node's behavior as seen in https://github.com/joyent/node/blob/6036e4f5a833b27fb474f3bf891c405606fdce19/lib/fs.js#L279-L285


rphillips:
nice. +1

kaustavha:
+1 I needs this
image

PR #780, opened 2015-07-23
fix(childprocess): do not prematurely close the handle

rphillips:


creationix:
So when does the user close it?

creationix:
+1

rphillips:
SIGCHLD will trigger the exit handler which calls :close()

PR #784, opened 2015-07-27
update querystring.lua

zhaozg:
Please think about this: name=asdf&desc=asdf&admin=asdf&telephone=asdf&mnode=1&mnode=2&mnode=3&mnode=4&cnode=1

sometimes we use same key to pass multi values,eg checkbox or select, This RP will make parse result to：  { desc = 'asdf', mnode = { '1', '2', '3', '4' }, admin = 'asdf',    telephone = 'asdf', cnode = '1', name = 'asdf' },

### DIFF
--- a/deps/querystring.lua
+++ b/deps/querystring.lua
@@ -62,7 +62,16 @@ function exports.parse(str, sep, eq)
     else
       local key, value = match(pair, '([^' .. eq .. ']*)' .. eq .. '(.*)')
       if key then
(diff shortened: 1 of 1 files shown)


creationix:
Looks good.  I can confirm this is the behavior in node as well.

```js
> require('querystring').parse("a=1&a=2&a=3")
{ a: [ '1', '2', '3' ] }
```

rphillips:
yeah, this is a nice fix. thanks!

PR #785, opened 2015-07-28
Replace print() with stdout:write() in repl.lua.

schwa423:
This allows the REPL to be hooked up to e.g. a JavaScript vt100 terminal via
WebSockets; without this change you can type into the terminal and each line
will be evaluated, but no output is sent back.

It then became trivial to hook up term.js to a WebSocket served by weblit-websocket.  This is the websocket handler I wrote (for example only, not production-quality):

```lua
local uv = require('uv')

-- This is called every time a new WebSocket client connects.
return function (req, read, write)
  stdout = {}
  function stdout:write (data)
    write({
      opcode = 1,
      payload = data
    })
  end

  stdin = { callback = nil, buffer = "" }
  function stdin:read_start (read_callback)
    self.callback = read_callback
    for i = 1, self.buffer:len() do
      self.callback(self.buffer:sub(i,i))
    end
    self.buffer = ""
  end
  function stdin:read_stop ()
    self.callback = nil
  end
  function stdin:set_mode (mode)
    -- TODO: don't know the intended semantics of set_mode...
    -- must read more about uv_tty_set_mode()
  end

  greeting = "Welcome to the REPL!!"
  function saveHistory (lines)
    print "saveHistory not implemented"
    write()
  end
  bogusInitialHistory = "these are the lines\nhere is a second line\n2 + 2 * 4\ninvokeFunction()"

  repl = require('repl')(stdin, stdout, greeting)
  repl.start(bogusInitialHistory, saveHistory)

  for message in read do
    if stdin.callback then
      stdin.callback(nil, message.payload)
    else
      stdin.buffer = stdin.buffer .. message.payload
      print("buffering: " .. stdin.buffer)
    end
  end

  print("about to do final write message ===========================?");
  write()
end
```

### DIFF
--- a/deps/repl.lua
+++ b/deps/repl.lua
@@ -49,7 +49,7 @@ setmetatable(exports, {
   })
   global._G = global
 
-  if greeting then print(greeting) end
+  if greeting then stdout:write(greeting) end
 
   local c = utils.color
 
@@ -62,14 +62,15 @@ setmetatable(exports, {
     for i = 1, results.n do
       results[i] = utils.dump(results[i])
     end
-    print(table.concat(results, '\t'))
+    stdout:write(table.concat(results, '\t'))
   end
 
   local buffer = ''
 
   local function evaluateLine(line)
+    stdout:write("\x1b[0G");
     if line == "<3" or line == "♥" then
-      print("I " .. c("err") .. "♥" .. c() .. " you too!")
+      stdout:write("I " .. c("err") .. "♥" .. c() .. " you too!")
       return '>'
     end
     local chunk  = buffer .. line
@@ -92,7 +93,7 @@ setmetatable(exports, {
         end
       else
         -- error
-        print(results[1])
+        stdout:write(results[1])
       end
     else
 
@@ -101,7 +102,7 @@ setmetatable(exports, {
         buffer = chunk .. '\n'
         return '>> '
       else
-        print(err)
+        stdout:write(err)
         buffer = ''
       end
     end
@@ -168,6 +169,8 @@ setmetatable(exports, {
       coroutine.wrap(function ()
         if line then
           prompt = evaluateLine(line)
+          stdout:write("\x1b[0G\n");
+          editor.position = 1
           editor:readLine(prompt, onLine)
           -- TODO: break out of >> with control+C
         elseif onSaveHistoryLines then
--- a/deps/repl.lua
+++ b/deps/repl.lua
@@ -62,15 +62,14 @@ setmetatable(exports, {
     for i = 1, results.n do
       results[i] = utils.dump(results[i])
     end
-    stdout:write(table.concat(results, '\t'))
+    stdout:write("\r" .. table.concat(results, '\t'))
   end
 
   local buffer = ''
 
   local function evaluateLine(line)
-    stdout:write("\x1b[0G");
     if line == "<3" or line == "♥" then
-      stdout:write("I " .. c("err") .. "♥" .. c() .. " you too!")
+      stdout:write("\rI " .. c("err") .. "♥" .. c() .. " you too!")
       return '>'
     end
     local chunk  = buffer .. line
@@ -93,7 +92,7 @@ setmetatable(exports, {
         end
       else
         -- error
-        stdout:write(results[1])
+        stdout:write("\r" .. string.gsub(results[1], "\n", "\n\r"))
       end
     else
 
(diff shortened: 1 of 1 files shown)


creationix:
Good catch.  Luvit already replaces lua's `print` with a wrapper around `stdout:write`, but since the repl allows custom stdout, we should use this directly instead of the print wrapper around the global stdout.

deps/repl.lua
```diff
@@ -49,7 +49,7 @@ setmetatable(exports, {
   })
   global._G = global
 
-  if greeting then print(greeting) end
+  if greeting then stdout:write(greeting) end
```
creationix: `print` adds a newline while `stdout:write` does not.  Doesn't this act weird when using the repl with normal stdout?

deps/repl.lua
```diff
@@ -62,14 +62,15 @@ setmetatable(exports, {
     for i = 1, results.n do
       results[i] = utils.dump(results[i])
     end
-    print(table.concat(results, '\t'))
+    stdout:write(table.concat(results, '\t'))
```
creationix: Same here, needs newline.

deps/repl.lua
```diff
@@ -62,14 +62,15 @@ setmetatable(exports, {
...
   end
 
   local buffer = ''
 
   local function evaluateLine(line)
+    stdout:write("\x1b[0G");
     if line == "<3" or line == "♥" then
-      print("I " .. c("err") .. "♥" .. c() .. " you too!")
+      stdout:write("I " .. c("err") .. "♥" .. c() .. " you too!")
```
creationix: Same here, needs newline.

deps/repl.lua
```diff
@@ -92,7 +93,7 @@ setmetatable(exports, {
         end
       else
         -- error
-        print(results[1])
+        stdout:write(results[1])
```
creationix: Same here, needs newline.

deps/repl.lua
```diff
@@ -101,7 +102,7 @@ setmetatable(exports, {
         buffer = chunk .. '\n'
         return '>> '
       else
-        print(err)
+        stdout:write(err)
```
creationix: Same here, needs newline.

deps/repl.lua
```diff
@@ -62,14 +62,15 @@ setmetatable(exports, {
...
       results[i] = utils.dump(results[i])
     end
-    print(table.concat(results, '\t'))
+    stdout:write(table.concat(results, '\t'))
   end
 
   local buffer = ''
 
   local function evaluateLine(line)
+    stdout:write("\x1b[0G");
```
creationix: What is this for?

schwa423: In all the places print() is replaced, adding a newline is insufficient, because the cursor isn't reset to the left (so each time you evaluate a line, the output and subsequent prompt keep moving to the right).  

I ganked this command string from readline.lua, where it has the comment "Cursor to left edge".  I'll look for a cleaner way to do this, now that I know that you're interested in the change in principle.

creationix:
If you want, you can add `"\r\n"` after each write and it will do what you expect.  The extra `\r` won't hurt modern terminals, but will help some older terminals that only move down but not left with a `\n` alone.

schwa423:
I missed your "\r\n" suggestion, but luckily great minds think alike :-p  

I ended up prepending "\r" instead; hopefully that's OK.  I also had to some find/replace for "\n" when an error triggers a stack trace.

zhaozg:
shoud be \r\n

creationix:
Good point about the stack traces being just `\n` internally.  Perhaps the best solution is to keep using unix style endings in core and do a stream replace in your websocket code since the `\r\n` is a local constraint of your client.  There will be other code also assuming unix line endings, so the only safe thing to do is to post-process the entire stream when you know the client terminal expects `\r\n` endings.

Could you change this PR to only replace `print(val)` with `stdout:write(val .. '\r\n')`.  That's all the luvit print was doing anyway. https://github.com/luvit/luvit/blob/master/deps/pretty-print.lua#L297

schwa423:
I'm pretty sure you mean `stdout:write(val .. '\n')` and not  `stdout:write(val .. '\r\n')`.

schwa423:
Good idea, using Unix line endings is a much better plan.

creationix:
New "luvit/repl" module is published to lit with your changes.  The next luvit release will include this change.

PR #786, opened 2015-07-30
http: Fix duplication of host/connection/user-agent request headers

squeek502:
Caused when the passed in headers have multiple keys, causing each `found` variable to depend on the iteration order of the headers table (each `found` var would only care about the last header iterated)

Example reproduction:

```lua
    local headers = { {"Front", "test"}, {"Host", "luvit.io"}, {"Behind", "test"} }
    local request = http.request({host="luvit.io", headers=headers})
    request:destroy()
    local num_hosts = 0
    for i,v in ipairs(request) do
      p(i, v)
      if v[1]:lower() == "host" then
        num_hosts = num_hosts + 1
      end
    end
    assert(num_hosts == 1)
```

### DIFF
--- a/deps/http.lua
+++ b/deps/http.lua
@@ -364,9 +364,9 @@ function ClientRequest:initialize(options, callback)
     self[#self + 1] = headers[i]
     local key, value = unpack(headers[i])
     local klower = key:lower()
-    host_found = host_found or (klower == 'host' and value)
-    connection_found = connection_found or (klower == 'connection' and value)
-    user_agent = user_agent or (klower == 'user-agent' and value)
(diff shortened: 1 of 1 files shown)


creationix:
Good catch.  Can we change the style too.

```lua
if not host_found and klower == 'host' then host_found = value end
```

The `if` should perform better.

creationix:
Since string comparisons are simple pointer comparisons in lua, we can probably remote the `not host_found and` prefix in the condition.  It will change the semantics slightly to match the last matching header instead of the first, but that's fine I think.

squeek502:
Done.

creationix:
Thanks!

PR #788, opened 2015-07-31
querystring: Add stringify

squeek502:
From the various commit messages:

- I changed urlencode to encode spaces as '%20' instead of '+'. This matches the behavior of Javascript's encodeURIComponent, which is what node.js uses.
- Implementation partially based on https://github.com/luvitrocks/luvit-querystring
- Node's implementation for reference: https://github.com/joyent/node/blob/master/lib/querystring.js

Note about the tests:

Iteration order of tables is arbitrary, so the stringify tests with multiple keys is a bit dubious whether or not they will always succeed, as the order could potentially be different.
I ordered the keys in the way that they are outputted on my system, but some safety might be required to make sure the tests succeed on every system

### DIFF
--- a/deps/querystring.lua
+++ b/deps/querystring.lua
@@ -50,6 +50,30 @@ function exports.urlencode(str)
   return str
 end
 
+local function stringifyPrimitive(v)
+  return tostring(v)
+end
+
+function exports.stringify(params, sep, eq)
+  if not sep then sep = '&' end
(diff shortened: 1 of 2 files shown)


creationix:
As long as the test are stable on the CI servers, I think this is fine.  I guess we'll see if people start seeing issues on other platforms / architectures when running the tests.

+1

PR #794, opened 2015-08-13
add test for ipv6 dns

rphillips:


rphillips:
travis does not support ipv6 :(

PR #795, opened 2015-08-18
add http-cookie support

zhaozg:
sometimes, we need build or parse http cookie header for Set-Cookie(response) or Cookie(request),

### DIFF
--- a/deps/http-cookie.lua
+++ b/deps/http-cookie.lua
@@ -1,0 +1,104 @@
+exports.name = "zhaozg/http-cookie"
+exports.version = "1.0.0-1"
+exports.homepage = "..."
+exports.description = "An adapter for build and parse http cookies."
+exports.tags = {"http", "cookie", "adapter"}
+exports.license = "MIT"
+exports.author = { name = "zhaozg" }
+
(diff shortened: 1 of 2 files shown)


zhaozg:
surprise, fail again

rphillips:
i kicked the builds again, since they 503'ed. passes. thanks!

creationix:
How do we want to draw the line between what gets included in core and what lives in lit as a third-party library?

With node, the decision was to draw the line between generic HTTP header parsing/encoding and cookie parsing/encoding.

Luvit is slightly different from node in that it's trivial to make a customized version of luvit with libraries added or removed.  Also building and testing luvit core libraries is no harder than other lit based libraries.  For these reasons there is less chance that a library will go stale once it becomes part of core.

zhaozg:
Sure. It's beter make cookie as lit. Please reset it, I'll try make it as lit package

creationix:
Ok, reset.  So are we deciding that we want to follow the lines node drew for deciding what goes into luvit core?

zhaozg:
Ok, follow the rule

PR #796, opened 2015-08-18
add luvit thread support.

zhaozg:
based on luv https://github.com/luvit/luv/commit/e536f14103f4a6abdda9a8b070bbc95f1511278d,
luvi https://github.com/luvit/luvi/commit/20142c2ce8be56b6bbb99d466ef87f27cd2356de,
and please see examples/multi-thread-tick.lua to know how to use

### DIFF
--- a/deps/thread.lua
+++ b/deps/thread.lua
@@ -1,0 +1,462 @@
+--[[
+
+Copyright 2014 The Luvit Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS-IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+--]]
+
+--- luvit thread management
+local uv = require('uv')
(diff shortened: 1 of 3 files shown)


deps/thread.lua
```diff
@@ -0,0 +1,464 @@
...
+--]]
+
+--- luvit thread management
+local uv = require('uv')
+local base = require('luvi').bundle.base
+
+exports.name = "luvit/thread"
+exports.version = "0.0.1"
+exports.dependencies = {
+}
```
creationix: No need for an empty table if there are no dependencies.  The field is optional.

rphillips:
can we add a test for this? or do we need a new luvi first?

creationix:
Why is the luvi logic inlined in this (bundle and path)?  Was this because you're waiting on a custom luvi build with the thread patch there too?

zhaozg:
Test need more time to code.please wait I to finish test code.

zhaozg:
@creationix I think thread should can use luvit deps lib. So need to use luvit require methods.

zhaozg:
Any simpler solution ?

zhaozg:
This need update luvi to work

creationix:
I thought the purpose of the luvi patch was to give threads access to the luvi environment without having to copy the luvi libraries?

creationix:
I found a way to clean this up.  It wasn't easy.

The luvi changes are in https://github.com/luvit/luvi/pull/118

And this PR needs the following changes https://github.com/luvit/luvit/commit/88280f938d64a4daf09b96bf4e27397dfb730c5d

creationix:
Moving to #799

PR #798, opened 2015-08-19
feat(childprocess/exec)

kaustavha:
First pass at integrating something like nodes exec which returns a callback with a buffer of all out&err data

### DIFF
--- a/deps/childprocess.lua
+++ b/deps/childprocess.lua
@@ -32,6 +32,8 @@ local core = require('core')
 local net = require('net')
 local timer = require('timer')
 local uv = require('uv')
+local los = require('los')
+local adapt = require('utils').adapt
 
 local Error = core.Error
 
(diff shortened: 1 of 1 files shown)


deps/childprocess.lua
```diff
@@ -164,4 +164,61 @@ local function spawn(command, args, options)
...
+  end
+  local function onClose(_exitCode)
+    exitCode = _exitCode
+    done()
+  end
+
+  local child = spawn(file, args, options)
+
+  child.stdout:on('data', function(chunk)
+    if #stdout > options.maxBuffer then
```
rphillips: you will need to keep track of the number of bytes read for each chunk type. #stdout isn't going to give you the right number.

rphillips:
Overall, we will need to use spawn() within the agent, since the overall output of a hostinfo might be over 200KB.

kaustavha:
yes ryan im just doing this for the lulz, i think the hostinfos look good with the steams and the run function atm.

rphillips:
The timeout support would be nice to have.

deps/childprocess.lua
```diff
@@ -164,4 +164,71 @@ local function spawn(command, args, options)
...
+  end
+
+  if options.shell then file = options.shell end
+  if not options.maxBuffer then options.maxBuffer = 200 * 1024 end
+
+  -- http://wowwiki.wikia.com/wiki/Lua_object_memory_sizes
+  -- We're being generous here since the internal representation is to chunk them into a table
+  -- which have a base size of 40bytes but the final callback recieves a single string of each type
+  local stdoutchunkbytes = 24
+  local stderrchunkbytes = 24
```
creationix: Why are we trying to take table size into account?  I would be very confused if I had set a max buffer size of 100 bytes and only got back 76 bytes.

deps/childprocess.lua
```diff
@@ -164,4 +164,71 @@ local function spawn(command, args, options)
   return em
 end
 
+local function exec(command, args, options, callback)
+  local file
+  local stringArgs = command..table.concat(args, ' ')
```
creationix: I think you're confusing `exec` with `execFile`.  In node, `exec` takes `(command, options, callback)` and runs the command in a subshell, but `execFile` takes `(file, args, options, callback)` and does not use a subshell.

deps/childprocess.lua
```diff
@@ -164,4 +164,71 @@ local function spawn(command, args, options)
...
+  if process.env._system_name:lower() == 'win32' then
```
rphillips: use los.type() here

deps/childprocess.lua
rphillips: also, we should chunk probably 4k at a minimum

deps/childprocess.lua
creationix: Also new luvi has `require('luvipath').isWindows` as a boolean.  This uses `jit.os == 'Windows'` on luajit and uses `not not package.path:match("\\")` as a fallback on other luas.

deps/childprocess.lua
kaustavha: You're right, but:
https://github.com/nodejs/node/blob/master/lib/child_process.js#L101-L107
So when I cycle back to this Ill port both

creationix: Yep, `execFile` is lower level and is used by `exec`.  The important thing I wanted to note is that `exec` doesn't take args, but does inject `sh` or `cmd` while `execFile` is just spawn with output buffered.

creationix:
You should add coroutine support to this to make it even easier to use from the repl.

See an example using the adapt helper at https://github.com/luvit/luvit/blob/master/deps/dns.lua#L696

creationix:
To use the `adapt` helper you'll need to split `execFile` into two functions.  One to be called directly by the end-user and an internal one that does the actual work, but requires all arguments (no optional slots).

The two user-facing functions will normalize inputs and then call the real function via `adapt`.  You'll need to check for both `"function"` and `"thread"` when looking for the `callback` after optional callbacks since either functions or coroutine threads are possible with adapt.

deps/childprocess.lua
```diff
@@ -164,4 +165,144 @@ local function spawn(command, args, options)
...
+    file = 'cmd.exe'
+    args = {'/s', '/c', '"'..stringArgs..'"' }
+  else
+    file = '/bin/sh'
+    args = {'-c', stringArgs }
+  end
+
+  if options and options.shell then file = options.shell end
+
+  return {
```
creationix: Just a style nit, but I prefer multiple return values here over creating a temporary table that will simple be destructured on the other side.  Also it seems you're never using the `cmd` property returned here and it's just one of the input parameters anyway, the caller already knows it.

deps/childprocess.lua
```diff
@@ -164,4 +165,144 @@ local function spawn(command, args, options)
...
+      table.insert(stderr, chunk)
+    end
+  end)
+
+  child:once('close', onClose)
+end
+
+local function exec(command, options, callback)
+  -- options is optional
+  local opts
```
creationix: If you modify `normalizeExecArgs` to return multiple values, this entire function can be simplified to:

```lua
return execFile(normalizeExecArgs(command, callback and options, callback or options)
```

`callback and options` will always be `options` if callback is true or falsy if callback is falsy.

creationix: Note that my proposed change preserves your logic.  Both will do the wrong thing if someone calls with an options table, but omits the callback.

deps/childprocess.lua
```diff
@@ -164,4 +166,142 @@ local function spawn(command, args, options)
...
+    callback = function() end
+    options = {}
+    args = {}
+  elseif not options then
+    options = {}
+  elseif not args then
+    args = {}
+  end
+
+  adapt(callback, _exec, file, args, options)
```
creationix: Need to tail call adapt for it to work (Return it's return value).

creationix: With a return added here, I can verify it works.

```lua
> require('childprocess').exec("ls", thread)
'ChangeLog\nLICENSE.txt\nMakefile\nNOTICE.txt\nREADME.markdown\nVagrantfile\nappveyor.yml\nbench\ndeps\nexamples\ninit.lua\nmain.lua\nmake.bat\npackage.lua\nprompt.lua\ntests\ntools\n'	''
> 
```

creationix:
+1

PR #800, opened 2015-08-20
in thread luvi.isWindows must be set before require('luvibundle')

zhaozg:


creationix:
Fixed this by moving `isWindows` check to luvipath module. https://github.com/luvit/luvi/pull/120/files  I don't want to have to duplicate the logic here.

PR #806, opened 2015-08-27
Makefile: support custom prefixes

bfontaine:


Makefile
```diff
@@ -5,6 +5,8 @@ LIT_VERSION=2.2.9
 LUVIT_TAG=$(shell git describe)
 LUVIT_ARCH=$(shell uname -s)_$(shell uname -m)
 
+PREFIX=/usr/local
```
rphillips: should be `PREFIX?=/usr/local` so the environment can override it

bfontaine: Fixed.

rphillips:
+1. Thanks!

PR #807, opened 2015-08-27
add threadpool support

zhaozg:


zhaozg:
continue test fail (it true fail), depend on make luv_work_ctx gc after all queued job done(https://github.com/luvit/luv/pull/183)

PR #811, opened 2015-10-05
fix(deps/dns.lua): resolv.conf parser enhancements

simonvetter:
This changes the parser so that it:
- ignores whitespaces at end of line
- treats lines starting with ';' as comments
- expects keywords to start at the beginning of the line.

From man resolv.conf:
Lines that contain a semicolon (;) or hash character (#) in the first column are treated as comments.
The keyword and value must appear on a single line, and the keyword (e.g., nameserver) must start the line.  The value follows the keyword, separated by white space.

### DIFF
--- a/tests/fixtures/resolve.conf.a
+++ b/tests/fixtures/resolve.conf.a
@@ -1,3 +1,8 @@
 ## Comment
+; Another comment
+notnameserver 2001:db8::a
 nameserver 192.168.0.1
+somegarbage
 nameserver ::1
+nameserver notanip
+nameserver ::2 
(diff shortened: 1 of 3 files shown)


tests/fixtures/resolve.conf.a
```diff
@@ -1,3 +1,8 @@
 ## Comment
+; Another comment
+notnameserver 2001:db8::a
 nameserver 192.168.0.1
-nameserver ::1
\ No newline at end of file
+somegarbage
+nameserver ::1
+nameserver notanip
+nameserver ::2
```
simonvetter: this line has an extra whitespace.

rphillips:
+1. thank you simon!

creationix:
Great improvement!

PR #813, opened 2015-10-07
fix(repl)

kaustavha:
Add path to the list of auto populated libs available to the user
My bad, guess I missed it somehow last time when I implemented the auto requires.


creationix:
Yep, I noticed this the other day, but was doing a presentation and forgot to take care of it later.

PR #822, opened 2015-10-29
Look in bundle first when doing module require

creationix:
This is changing the require behavior to be less powerful, but also less confusing.

Before if you had modules in your filesystem that matched/conflicted with names in the bundle, it could load either depending on your base. (Often in unexpected ways)

With the change, the internal version will always be preferred.


rphillips:
+1

PR #842, opened 2015-11-24
Enable --256-colors if $TERM = *-256color

urzds:
This patch enables 256 colours for all terminals claiming to support it (e.g. TERM=screen-256color)


rphillips:
Thanks!!

PR #848, opened 2015-12-01
upgrade to support newest lua-openssl

zhaozg:


creationix:
This requires the new ssl bindings correct?  I see it failed here using the old ones.

zhaozg:
yes，please update submodule lua-openssl with newest master branch

PR #849, opened 2015-12-01
remove broken example file fs-coroutines.lua, fs-watchfile.lua, speed-http.lua, test-http-ugly.lua

zhaozg:


rphillips:
does this need a new luvi and lua-openssl upgrade?

zhaozg:
Yes,lua-openssl  need upgrade

PR #850, opened 2015-12-01
tap: check for closing handles in shutdown

### LINKED ISSUE #839: luvit/tap - when calling tap() from a timer: Uncaught Error: handle 0x403bd828 is already closing
The following code causes the error below:
```lua
local tap = require("tap")
local uv = require("uv")

local function run_tests()
  local req = uv.fs_scandir("tests")

  repeat
    local ent = uv.fs_scandir_next(req)

    if not ent then
      -- run the tests!
      tap(true)
    end
    local match = string.match(ent.name, "^test%-(.*).lua$")
    if match then
      local path = "./test-" .. match
      tap(match)
      require(path)
    end
  until not ent
end

local timer = uv.new_timer()

local function wait()
  run_tests()
end

timer:start(1000, 0, wait)
```

```
Uncaught Error: handle 0x403bd828 is already closing
stack traceback:
        [C]: at 0x004e1b30
        [C]: in function 'walk'
        [string ".../libs/tap.lua"]:131: in function 'run'
        [string ".../libs/tap.lua"]:163: in function 'tap'
        [string ".../tests/debu..."]:12: in function 'run_tests'
        [string ".../tests/debu..."]:26: in function <[string ".../tests/debu..."]:25>
        [C]: in function 'run'
        [string "bundle:init.lua"]:52: in function <[string "bundle:init.lua"]:47>
        [C]: in function 'xpcall'
        [string "bundle:init.lua"]:47: in function 'fn'
        [string "bundle:deps/require.lua"]:293: in function <[string "bundle:deps/require.lua"]:258>
```

The code above is a MWE / shortened testcase.

In my actual code I was trying to start a docker container and wait for the HTTP service inside to become available, testing it using `http.get()` calls executed from a timer.

rphillips:
fix #839

### DIFF
--- a/tests/libs/tap.lua
+++ b/tests/libs/tap.lua
@@ -128,7 +128,9 @@ local function run()
   end
 
   -- Close all then handles, including stdout
-  uv.walk(uv.close)
+  uv.walk(function(handle)
+    if not uv.is_closing(handle) then uv.close(handle) end
+  end)
   uv.run()
 
   os.exit(-failed)


creationix:
Good catch, I thought I had all these updated some time ago.

+1

PR #858, opened 2015-12-13
When requiring a module, allow it to mutate the global namespace.

### LINKED ISSUE #857: require fallback doesn't seem to work
> [11:40:20]  <laarc>	Here's an easy repro case for the luvit require problem I'm running into...
> [11:40:34]  <laarc>	git clone https://github.com/sctb/lumen
> [11:40:37]  <laarc>	cd lumen
> [11:40:46]  <laarc>	LUMEN_HOST=luvit bin/lumen
> [11:41:17]  <laarc>	That will blow up with an error:  No such module 'compiler' in ...
> [11:41:53]  <laarc>	But LUA_PATH is set at that point
> [11:43:22]  <laarc>	so it unfortunately doesn't seem true that luvit falls back to lua's normal require whenever luvit can't find something

laarc:
Closes #857.

This PR modifies `luvit/deps/require.lua`:  All modules now share a global namespace. 

Consider the following module `foo.lua`:

```lua
x = 42
```

And the following module `bar.lua`:

```lua
require("foo")
print(x)
```

Now consider `require("bar")`.  Using standard Lua or LuaJIT, this causes `foo.lua` to be required, which sets `_G.x` to `42`, then ultimately prints `42` to stdout.

However, using Luvit's `require`, no mutations to `_G` were allowed.  Therefore `print(x)` would print `nil`.

I understand that there may be an argument for isolating each module in this manner.  Unfortunately, [Lumen](https://github.com/sctb/lumen) is designed to rely on the fact that requiring a module mutates the global state.  Since I'm trying to add `luvit` support to Lumen, this puts me in something of a tricky position.

I was hoping that you'd be willing to accept this PR (which is all that's needed for me to be able to support `luvit` in Lumen).  

For what it's worth, this PR does not break any tests in the test suite.  `make test` results in "All tests passed."

Either way, thanks for considering this.  Have a great week!

### DIFF
--- a/deps/require.lua
+++ b/deps/require.lua
@@ -289,7 +289,7 @@ function Module:require(name)
         return module:require(...)
       end
     }
-    setfenv(fn, setmetatable(global, { __index = _G, __newindex = _G }))
+    setfenv(fn, setmetatable(global, { __index = _G, __newindex = _G })) 
     local ret = fn()
 
     -- Allow returning the exports as well
--- a/deps/require.lua
+++ b/deps/require.lua
@@ -289,7 +289,7 @@ function Module:require(name)
         return module:require(...)
       end
     }
-    setfenv(fn, setmetatable(global, { __index = _G }))
+    setfenv(fn, setmetatable(global, { __index = _G, __newindex = _G }))
     local ret = fn()
 
     -- Allow returning the exports as well


laarc:
When travis-ci ran the tests against my original commit, the `resolve4` dns test failed:  https://travis-ci.org/luvit/luvit/jobs/96620357

Not sure why.  It doesn't seem related, though. I pushed another commit in order to cause travis-ci to re-run the tests, and this time all of them passed.

creationix:
I think this was fixed already by https://github.com/luvit/luvit/commit/5e5d0833acc92e911813c2c71c427b51df8bb5fb where I had a fake `_G`.  Try with luvit master and see if that helps.

laarc:
That was necessary, but not sufficient. :)  This PR is required since it enables luvit modules to mutate the global namespace during a `require`.   Lumen's design unfortunately requires this, since it relies on Lua's standard behavior of allowing global mutations.

I was hoping I'd be able to avoid forking luvit...  `brew install luvit ; LUMEN_HOST=luvit lumen` is the ideal scenario for Lumen's users.

laarc:
Also, if you're wary of making this PR request the default luvit behavior, maybe it could be optional instead?  i.e. toggled on via an environment variable?  That'd work.

creationix:
I don't want luvit modules to automatically create globals if you forget a `local`.  This is considered bad practice.  But I'm fine with allowing global mutation with _G. prefix.  Does explicit global creation not work using `_G` with my patch?

laarc:
Unfortunately, the situation I'm in is that the author of Lumen has control over Lumen, not me.  He feels a certain way about the aesthetics, and _G turns out to be very problematic for a variety of absurd reasons.  If you'll humor me for a moment, I'd love to try to explain why:

https://github.com/sctb/lumen/blob/master/bin/compiler.lua#L107-L137

The author relies on Lua's standard behavior:  That function isn't declared local, so therefore it's exposed globally.  This is necessary for the proper operation of the compiler.

I get that _G is a workaround.  But I implore you not to make this design decision.  The author of Lumen wrote a compiler which generates a `compiler.lua` file that conforms to Lua's standard behavior.  There is roughly zero chance that I'll be able to persuade the author to modify their compiler to generate `_G.bind = function bind(lh, rh)` rather than `function bind (lh, rh)` for toplevel forms. 

I see the engineering tradeoff, but it seems like the author would be correct in this case.  Forget about aesthetics.  Every new Lumen user would wonder why Lumen has to generate something as complicated as `_G.bind = function bind(lh, rh)` rather than `function bind(lh, rh)`.   Even if we set aside the question of whether it's pretty, that non-standard complexity means the compiler will be forced to become more complex; it will have to incorporate some surprising code into the codebase which Lumen's contributors will need to maintain and document.

Here's how simple Lumen's code is right now.  The logic for compiling a function into either javascript or lua is *three* lines of code in total: 

https://github.com/sctb/lumen/blob/master/compiler.l#L383-L385

```
    (if (= target 'js)
        (cat "function " id args " {\n" body ind "}" tr)
      (cat p "function " id args "\n" body ind tr))))
```

The first line checks whether the user wants Lumen to compile to javascript, or to lua.  The second line generates a javascript-style function.  The third line generates a Lua-style function.

And the cool thing is, whether Lumen is running via node (`bin/lumen.js`) or whether it's running via lua (`bin/lumen.lua`), both of them can compile Lumen's codebase at runtime into either javascript or lua!   It's like running in two languages at once.  Quite an impressive feat of engineering..

Here's where the compiled function is then bound to a global variable:

https://github.com/sctb/lumen/blob/master/compiler.l#L612-L613

```
(define-special %global-function (name args body) :stmt :tr
  (if (= target 'lua)
      (let x (compile-function args body name: name)
        (cat (indentation) x))
    (compile `(set ,name (%function ,args ,body)) :stmt)))
```

To incorporate your `_G` workaround for toplevel function definitions, the code would have to be changed significantly, and the aesthetics of the result won't line up with the author's sense of design.

All of that is to explain I have zero chance of modifying Lumen's codebase to use `_G` in this case.  And at that point, my options become:  Fork Lumen and insert the luvit-specific `_G` workaround, or fork `luvit` and add this pull request to it (which is to say, make `luvit` conform to Lua's standard behavior).  Which would you choose?

The original motive for disallowing global mutation seems to be to protect people.  That's commendable, but consider that (a) very few people have been bitten by the bug you're trying to protect against, and (b) by forcing non-standard behavior, third parties are forced to write code that targets luvit specifically, rather than lua generally.

I also get that I'm being something of a pain.  I apologize for that.  Trying to grow a programming language from zero users is Really Hard.  The more implementations that Lumen runs on, the higher the chances of success.  But if I have to fork luvit in order to force it to conform to Lua's standard behavior, then that means Lumen users will never be able to do `brew install luvit`.  Worse, they'll probably do `brew install luvit` and then wonder why `LUMEN_HOST=luvit lumen` blows up with an error.  That's precisely the type of bad first impression that I'm trying very hard to avoid.  (It took three days to diagnose this problem and about an hour to get this pull request together, for what it's worth.)

I would appreciate very much, and would owe you greatly, if there would be any way of getting this pull request's functionality into Luvit.  This is last remaining obstacle standing in the way of `luvit` support for Lumen.

There are a variety of ways to add this to luvit without making it the default behavior.  It could be optional at runtime:  Luvit could check for whether a special environment variable is set, for example.  A new command-line argument could be added.   Or `require.lua`'s initialization function could take a new parameter, `allow_global_mutations`.

Any of those would work. `_G` happens to be the one thing that won't.

I wish I could show some actual *results* to demonstrate why anyone should care about this... I'm imagining how all of this must sound from your point of view.  I think you'd be totally justified in telling me "no, sorry." 

All I can do is be honest about why this is a tricky situation, then hope for the best.

Regardless of how this turns out, thanks for your time, and for all the software you've written over the years.  Your `uv_run` module for Node is something that recently helped me a lot, and it was really cool to discover that you wrote it.

daurnimator:
> I get that _G is a workaround. But I implore you not to make this design decision. The author of Lumen wrote a compiler which generates a compiler.lua file that conforms to Lua's standard behavior. 

It's considered poor lua form to add globals.
The Luman author should use local variable for this sort of thing; or export a table.

> Every new Lumen user would wonder why Lumen has to generate something as complicated as _G.bind = function(lh, rh) rather than function bind(lh, rh)

Why is it making globals in the first place? Why isn't it making locals?
i.e.`local function bind(lh, rh)`
Or if you need to export, either `function exports.bind(lh, rh)` or `local function bind(lh, rh) ... end; exports.bind = bind`

creationix:
@laarc As @daurnimator has mentioned, you can make a shim that creates the environment lumen expects while still using stock luvit.

I'm more concerned with the basic workflow you're imagining.  Installing luvit via homebrew doesn't really work since their update system is incompatible with luvit's.  Must you use homebrew, or is installing luvit via the normal channels an option for your users?

laarc:
Circa 2016, OS X users generally expect programming tools to be available via Homebrew.  It's not really a question of whether that's good or bad.  It's simply the world right now.

FWIW, the motive for using luvit in Lumen is that it exposes a wide range of very useful libraries.  Everything from DNS querying, to the `uv` module, to the `fs` module.  (The `fs` module is really a killer feature.)

Therefore, if this PR (or some optional variant) is merged, it's a matter of time until `brew install luvit` comes with this functionality.  At that point, whether it's three months or three years from now, `brew install luvit` will install a version of luvit that works with Lumen.

I'm building a language on top of Lumen, and right now it builds a custom version of `luvit` that includes this PR.  But I would really like to get away from this hack, moving forward, and use a stock `luvit` installation.

> You can make a shim that creates the environment lumen expects while still using stock luvit.

Where?  There's no opportunity to create one on startup.  A concrete suggestion would be helpful.  (I  struggled to do what you suggested for several hours with no success, just now.)

Or, alternatively, you can try it yourself.  Again, here are the step-by-step instructions:

```
git clone https://github.com/sctb/lumen
git clone https://github.com/luvit/luvit
cd luvit
make
cd ../lumen
LUMEN_HOST=`pwd`/../luvit/luvit bin/lumen
```

You can literally copy-paste that into your terminal.  Lumen dies with an error on startup.  What sort of shim are you saying to use?

laarc:
Ugh.  I finally figured out the magical hack to get everything to work properly.

Hello, people from the future.  You have no doubt arrived here from Google, and you're looking for a way to make Luvit's `require` function behave identically to Lua's, yet still be able to call `require("pretty-print")`.  You're in luck, I have just the thing.

Put this at the top of your main .lua file.  (Whichever .lua file you're passing in to `luvit` on the command line.)

```lua
_G.luvit_require = require
_G.lua_require = _G.require
local global = {}
setmetatable(global, {__index = _G, __newindex = _G})
setfenv(1, global)
require = _G.luvit_require
```

Getting to this point stole >20 hours of my time.  (Cue the world's tiniest violin to start playing.)

@creationix, are you certain you want me to try to persuade the author of Lumen to include this hack?  I would really, really prefer not doing that.  Lumen's author is extremely busy, and I'm trying to bother them only when I have absolutely no other choice.

Would you be willing to add a command-line switch to `luvit` to automatically do this on startup?  I can submit that as a new pull request if you want, but what are your thoughts about the idea?

Closing issue #857, PR #858.

creationix:
@laarc, I'm so sorry I didn't see your updates here.  I've been extremely busy lately with family stuff.

To answer your question, yes I want to fix luvit to play nice.  I've even gone so far as to propose replacing luvit's require entirely with a couple lua loaders and use lua's native require directly.  This new loader will use `loadfile` with no environment hacks at all.

Also I recently fixed a bug in lit's zip generating logic that was one of the blockers preventing a proper homebrew install of luvit.  The only remaining issue it we'll have to disable lit's update feature if installed via homebrew and instead print a message telling the user to use homebrew's update system instead.

https://github.com/luvit/luvit/blob/luvly/luvit-loader.lua - prototype of require replacement.
https://github.com/luvit/lit/issues/115 - lit's update issue with homebrew

creationix:
Made more progress with the luvly branch, see this PR https://github.com/luvit/lit/pull/155

PR #861, opened 2016-01-06
RFR: Switch luvit modules to luvit-loader format

creationix:
Initially this had grand ambitions to remove `luvit/require` from luvit.  But that is a breaking change for many people using luvit.  Now this is a simple change to remove dependence on the global `exports` table so that luvit modules are more portable for luv users.

We'll still need to wean off of the luvi libraries and current openssl bindings for full portability, but this is a great start and many of the simpler modules here will now be portable in plain lua applications as-is.


creationix:
Also this PR introduces a new "resource" module that can replace the global `module.*` methods used for relative resource loading.

Issue #872, opened 2016-02-02
Add windows support to look up DNS resolvers.


creationix:
Currently we only look up servers for systems with `resolve.conf`.  We should add support to FFI into the windows APIs to look this up as well.

This needs to be added to `loadResolvers` in the dns module https://github.com/luvit/luvit/blob/b382c9ba6d3057f900a20bdd41e4feceedd13145/deps/dns.lua#L744-L785

Also luvit's init function needs to be changed to call this unconditionally once this feature is added. https://github.com/luvit/luvit/blob/b382c9ba6d3057f900a20bdd41e4feceedd13145/init.lua#L34

PR #873, opened 2016-02-02
Fix: use system setting for dns server addresses

rjemanuele:
This was missing from Windows

Fixes https://github.com/luvit/luvit/issues/872

### DIFF
--- a/deps/dns.lua
+++ b/deps/dns.lua
@@ -20,7 +20,7 @@ limitations under the License.
 -- https://github.com/openresty/lua-resty-dns/blob/master/lib/resty/dns/resolver.lua
 
 exports.name = "luvit/dns"
-exports.version = "1.0.2"
+exports.version = "1.1.0"
 exports.dependencies = {
   "luvit/dgram@1.1.0",
   "luvit/fs@1.2.2",
@@ -41,6 +41,7 @@ local net = require('net')
 local timer = require('timer')
 local Error = require('core').Error
 local adapt = require('utils').adapt
+local ffi = require('ffi')
 
 local bit = require('bit')
 local crypto = require('tls/lcrypto')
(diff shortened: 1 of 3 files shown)


deps/dns.lua
```diff
@@ -20,7 +20,7 @@ limitations under the License.
 -- https://github.com/openresty/lua-resty-dns/blob/master/lib/resty/dns/resolver.lua
 
 exports.name = "luvit/dns"
-exports.version = "1.0.2"
+exports.version = "1.0.3"
```
creationix: Since this has added functionality, but is still backwards compatible, we should bump the middle version number

rjemanuele: Ok, will do.

deps/dns.lua
```diff
@@ -741,7 +742,89 @@ exports.setDefaultServers = function()
...
+  local servers = {}
+
+  ffi.cdef[[
+    typedef uint32_t DWORD; //Integer
+    typedef uint32_t ULONG; //Alias
+    typedef uint32_t UINT; //Alias
+    typedef ULONG *PULONG; //Pointer
+
+    enum {
+      ERROR_SUCCESS         = 0L,
```
rjemanuele: I am wondering why this is reported as a "redefine".

deps/dns.lua
```diff
@@ -741,7 +742,89 @@ exports.setDefaultServers = function()
...
+      MAX_HOSTNAME_LEN      = 128,
+      MAX_DOMAIN_NAME_LEN   = 128,
+      MAX_SCOPE_ID_LEN      = 256
+    };
+
+    typedef struct {
+      char String[16];
+    } IP_ADDRESS_STRING, *PIP_ADDRESS_STRING, IP_MASK_STRING, *PIP_MASK_STRING;
+
+    typedef struct _IP_ADDR_STRING {
```
rjemanuele: Same here, redefine for _IP_ADDR_STRING

rphillips:
+1! nice!

PR #875, opened 2016-02-06
bump to luvi 1.8.0-3 with fs_scandir_next API change

rphillips:
needs the new luvi release

### DIFF
--- a/deps/fs.lua
+++ b/deps/fs.lua
@@ -182,9 +182,10 @@ local function readdir(path, callback)
     local files = {}
     local i = 1
     while true do
-      local name = uv.fs_scandir_next(req)
+      local name, typ = uv.fs_scandir_next(req)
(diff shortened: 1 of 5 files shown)


deps/fs.lua
```diff
@@ -180,9 +180,9 @@ local function readdir(path, callback)
     local files = {}
     local i = 1
     while true do
-      local ent = uv.fs_scandir_next(req)
-      if not ent then break end
-      files[i] = ent.name
+      local name = uv.fs_scandir_next(req)
+      if not name then break end
+      files[i] = name
```
creationix: The only problem here is lit packages have no way to signify what versions of luv they depend on.  If we publish a new fs that uses the new style from luvi 2.6.0, then anyone using an older version will get broken.

PR #880, opened 2016-03-18
change tcp-echo-server to node-style api

Miigon:
easier and clearer :)

### DIFF
--- a/examples/tcp-echo-server.lua
+++ b/examples/tcp-echo-server.lua
@@ -1,32 +1,9 @@
-local uv = require('uv')
+local net = require ("net")
 
--- Create listener socket
-local server = uv.new_tcp()
-server:bind('127.0.0.1', 1234)
-
-server:listen(128, function(err)
-  -- Create socket handle for client
-  local client = uv.new_tcp()
-
-  -- Accept incoming connection
-  server:accept(client)
-  print("Client connected")
-
-  -- Relay data back to client
-  client:read_start(function(err, data)
-    -- If error, print and close connection
-    if err then
-      print("Client read error: " .. err)
-      client:close()
-    end
-
-    -- If data is set the client has sent data, if unset the client has disconnected
-    if data then
-      client:write(data)
-    else
-      print("Client disconnected")
-      client:close()
-    end
-
+local server = net.createServer(function(client)
+  client:on("data", function(data)
+    client:write(data)
   end)
 end)
+
+server:listen(1234, "127.0.0.1")


creationix:
I like having an example that shows how to use libuv directly.  Also now that I look closer, both examples have errors.  Ideally we should have 3 examples, one showing the node style correctly, one showing the coroutine style correctly and one showing raw uv style directly.

examples/tcp-echo-server.lua
```diff
@@ -1,32 +1,9 @@
...
-      print("Client read error: " .. err)
-      client:close()
-    end
-
-    -- If data is set the client has sent data, if unset the client has disconnected
-    if data then
-      client:write(data)
-    else
-      print("Client disconnected")
-      client:close()
```
creationix: This was wrong, we should call shutdown first to flush out any data and then close in it's callback.

Also we may need to check if it's already closing before calling close.

examples/tcp-echo-server.lua
```diff
@@ -1,32 +1,9 @@
...
-  client:read_start(function(err, data)
-    -- If error, print and close connection
-    if err then
-      print("Client read error: " .. err)
-      client:close()
-    end
-
-    -- If data is set the client has sent data, if unset the client has disconnected
-    if data then
-      client:write(data)
```
creationix: There is no back-pressure here, we should read_stop if the write call backs up and then read_start again when the buffer is cleared.

examples/tcp-echo-server.lua
```diff
@@ -1,32 +1,9 @@
...
-    else
-      print("Client disconnected")
-      client:close()
-    end
+local net = require ("net")
 
+local server = net.createServer(function(client)
+  client:on("data", function(data)
```
creationix: I think the node style has a `:pipe()` method that could make this even easier.

```lua
client:pipe(client)
```

Otherwise we need to also handle shutdown and close properly as is done inside pipe.

creationix:
If you have time to research the proper way to do all three styles and include them in the PR, that would be awesome.  I'm pretty busy with stuff for the next few days, but I can answer questions and point you in the right direction.

Miigon:
I admit that coroutine is more directly to express the logic of server.But an echo server based on coroutine may be make a big fuss over a minor issue?I mean,echo server just relay the data back and doesn't have any logic.

creationix:
Once you fully implement an echo server with proper back-pressure error handling and stream termination, you'll see that it's not quite as simple as imagined.  Yes node makes it appear simple with the pipe helper (I helped design this in node back in 2010), but that's just a high-level tool designed explicitly to make echo server a 1-liner.

An example using `:pipe` is neat to show how easy it is when you use code already written for this, but it isn't very instructive about how to handle the various events and use cases for real apps.

I would love examples showing all the various possible techniques so it's easier to compare strengths, even such a simple server as echo.

Miigon:
Yeah,you are right.
I wrote node style and rawuv style echoserver just now.
And now writing the coroutine style one.
How should it be like?
Use a function to do the all things during the connection alive?

creationix:
For coroutine style, use the coro-net library.  There are lots of examples of using it in the lit code, weblit code and wscat utility.
https://github.com/creationix/wscat
https://github.com/creationix/weblit
https://github.com/luvit/lit

creationix:
For an echo server, the main loop would look like:


```lua
for chunk in read do -- read will block will there is a chunk on the socket
  write(chunk)
end
-- When EOS of found, the loop will terminate
write() -- writing nil closes our end of the socket.
```

PR #899, opened 2016-05-19
repl: Fix error() calls with a non-string message breaking the repl

### LINKED ISSUE #898: error() quits interactive luvit, but error("hi there") does not.
Inconsistent behavior of `error()` in luvit.  Is this expected behavior?  I would not expect this to be the case:

```
> function foo()
>> error()
>> end
> foo()
Uncaught Error: [string "bundle:deps/repl.lua"]:183: [string "bundle:deps/repl.lua"]:102: attempt to concatenate a nil value
stack traceback:
	[builtin#36]: at 0x010b6e0b30
	[string "bundle:deps/repl.lua"]:183: in function <[string "bundle:deps/repl.lua"]:173>
	[C]: in function 'run'
	[string "bundle:init.lua"]:52: in function <[string "bundle:init.lua"]:47>
	[C]: in function 'xpcall'
	[string "bundle:init.lua"]:47: in function 'fn'
	[string "bundle:deps/require.lua"]:300: in function <[string "bundle:deps/require.lua"]:265>
MQD2ABG8WL:luvit sam5373$ function foo() begin
-bash: syntax error near unexpected token `begin'
MQD2ABG8WL:luvit sam5373$ luvit
Welcome to the Luvit repl!
> function foo()
>> error("Hi there")
>> end
> foo()
[string "REPL"]:2: Hi there
stack traceback:
	[C]: in function 'error'
	[string "REPL"]:2: in function <[string "REPL"]:1>
	[C]: in function 'xpcall'
	[string "bundle:deps/repl.lua"]:93: in function 'evaluateLine'
	[string "bundle:deps/repl.lua"]:177: in function <[string "bundle:deps/repl.lua"]:175>
> _
```

squeek502:
Fixes #898

`error()` calls with non-string messages will now output the following:

```
Welcome to the Luvit repl!
> error(function() end)
error with unexpected error message type (function), no stacktrace available
> error()
error with unexpected error message type (nil), no stacktrace available
```

Let me know if you think there's a better way to handle this, or if the message could be made clearer.

### DIFF
--- a/deps/repl.lua
+++ b/deps/repl.lua
@@ -97,9 +97,12 @@ return function (stdin, stdout, greeting)
         if results.n > 0 then
           printResults(results)
         end
-      else
+      elseif type(results[1]) == 'string' then
         -- error
         stdout:write(results[1] .. '\n')
+      else
+        -- error calls with non-string message objects will pass through debug.traceback without a stacktrace added
+        stdout:write('error with unexpected error message type (' .. type(results[1]) .. '), no stacktrace available\n')
       end
     else
 


creationix:
Maybe instead of the type, run the value through `tostring`?

squeek502:
Using `tostring` would be largely the same except in the case of `boolean`, where the output would be something like `error with unexpected error message type (false), no stacktrace available`, which seems like it could be confusing.

I'm not sure there's much benefit to displaying the value, but I'll leave it up to you.

For reference, here's what the plain Lua 5.1.5 repl outputs (note that `error()` is totally silent):

```
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
> error()
> error(false)
(error object is not a string)
> error(function() end)
(error object is not a string)
```

PR #906, opened 2016-06-20
Changed random seed from os.clock to os.time

SinisterRectus:
Luvit currently uses `os.clock()`, a measure of processor time, to seed the RNG, but this value is often not unique. `os.time()` guarantees a unique seed as long as at least one second lapses between executions, and is the [recommended method](https://www.lua.org/pil/18.html).


rphillips:
thanks!

PR #912, opened 2016-06-23
dns: fix server order resolving

rphillips:
The order of resolver preferences were not being honored

A `resolve.conf` with the following config would fail and use the last nameserver in the list first - assuming the 192.168.0.1 is a working nameserver:

```
nameserver 192.168.0.1
nameserver 8.8.8.8
```

The expected result would be for the 8.8.8.8 server to be queries first.

Note: The luvit test runner does not pull in the latest sources within the tree.


creationix:
+1

PR #916, opened 2016-06-24
Move http.headerMeta to its own package and add util functions

squeek502:
Adds luvit/http-header package containing:
- headerMeta (moved from http)
- newHeaders() (returns a new table with headerMeta as its metatable)
- toHeaders(tbl) (converts a table of header values into a table with headerMeta as its metatable)
- combineHeaders(...) (converts and combines tables into a single table with headerMeta as its metatable)
- getHeaders(tbl) (extracts headers from the array-like portion of tbl into a new table with headerMeta as its metatable)

---

See the comments in http-header.lua for slightly more info on each function.

**Rationale:**

http.headerMeta can be useful outside of the Luvit `http` package. For example, it (along with the added utility functions) could be used with `coro-http` like so:

```lua
local http = require('coro-http')
local httpHeader = require('http-header')

local headers = httpHeader.newHeaders()
headers['accept'] = 'application/json'

local res, body = http.request('GET', 'https://lit.luvit.io', headers)

local resHeaders = httpHeader.getHeaders(res)
assert(resHeaders['content-type'] == 'application/json')
```

**Notes**

- Let me know if you'd like any changes to the name of the package, the names of the functions, and/or any of the comments--I definitely think they can be improved.


creationix:
Maybe `toHeaders` could be `combineHeaders`?

squeek502:
I originally had it as combineHeaders, but I thought that might be slightly misnamed for the potentially-common use case of passing it a single table to convert. For example:

```lua
local httpHeader = require('http-header')

local headersKV = {
    ['Host'] = "",
    ['Accept'] = "",
    ['Content-Length'] = 0
}

local headers = httpHeader.toHeaders(headersKV)
```

creationix:
Maybe split it out into two functions then?  The inner loop could be `toHeaders` I think.

squeek502:
That might make the usage of `toHeaders` less apparent, because it seems like it'd need to have a second optional parameter to be used by `combineHeaders`. Example:

```lua
local function toHeaders(tbl, headersTable)
  headersTable = headersTable or newHeaders()
  if tbl then
    for k,v in pairs(tbl) do
      if type(k) == "number" then k = #headersTable + 1 end
      headersTable[k] = v
    end
  end
  return headersTable
end

local function combineHeaders(...)
  local combined = newHeaders()
  for _, tbl in ipairs({...}) do
    toHeaders(tbl, combined)
  end
  return combined
end
```

Let me know if you can think of a better implementation

squeek502:
Split `toHeaders(...)` into `combineHeaders(...)` and `toHeaders(tbl)`  using a shared function that is not returned from the file, as I'm not sure if it's worth exposing. The function in question:

```lua
-- Adds all header information found in `tbl` into `headers`, which should be a table
-- with headerMeta as its metatable.
--
-- Note: String keys in `tbl` will overwrite the key's value(s) in `headers` if it exists
local function appendToHeaders(tbl, headers)
    if tbl then
      for k,v in pairs(tbl) do
        if type(k) == "number" then k = #headers + 1 end
        headers[k] = v
      end
    end
    return headers
end
```

creationix:
+1

squeek502:
Squashed and added `http-header` to `package.lua`.

squeek502:
Linux CI errored due to https://github.com/luvit/lit/issues/184, will be fixed once Lit updates.

creationix:
I think if you update the lit version in the various makefiles, it will fix the build problem.

squeek502:
`get-lit.sh` is still pulling v3.4.1

creationix:
I see, you can update those as well if you want.

squeek502:
Should build successfully once luvit/lit#189 is merged.

squeek502:
Build still failing after luvit/lit#189 because the lit 3.4.2 tag still points to to commit before luvit/lit#189. Need to either move the tag or release a new lit version.

creationix:
I moved the tag and restarted the travis build.

*fingers crossed*

creationix:
Sorry for all the trouble and thanks a million for finally fixing this bug in lit!

squeek502:
No problem!

PR #917, opened 2016-06-25
Add os.time() to random seed

LoganDark2:
`os.time` by itself is not very accurate, yet `os.clock()` is not very unique. The solution? Multiply them together! This will almost always generate guaranteed random numbers.


rphillips:
os.time() is more correct to seed Lua's RNG.

LoganDark2:
well os.clock is accurate in 1/100 of a second

Issue #843, opened 2015-11-24
How to use Luvit with LuaCov or LuaTrace?


urzds:
I am trying to make a coverage analysis of my tests using [LuaCov](http://keplerproject.github.io/luacov/). However, the coverage stats only include measures of `luvipath.lua` and `luvibundle.lua`, but completely miss anything else.

Example with `testthis.lua`:
```lua
function testthis()
        print("testthis")
end

testthis()
```

```
# luvit -e 'jit.off(); jit.flush(); local luacov = require("luacov");' testthis.lua
testthis
# cat luacov.stats.out
107:/home/tim/luvi/src/lua/luvipath.lua
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 11 0 0 11 31 20 0 11 0 0 5 2 0 3 0 0 0 0 5 5 0 0 0 5 5 11 22 11 6 0 0 0 5 4 0 0 0 5 16 11 31 20 0 11 0 0 0 5 5 25 20 20 20 1 19 1 0 18 0 0 0 0 0 5 13 8 8 0 0 5 5
```

I compared this with plain LuaJIT:
```
# rm luacov.stats.out
# luajit -e 'jit.off(); jit.flush(); local luacov = require("luacov");' testthis.lua
testthis
# cat luacov.stats.out
5:testthis.lua
1 1 2 0 2
```

At first I assumed that all of this is caused by LuaJIT, as [described for LuaTrace](https://github.com/geoffleyland/luatrace/blob/master/README.md):
> luatrace runs under "plain" Lua and LuaJIT with the -joff option (LuaJIT doesn't call hooks in compiled code, and luatrace loses track of where it's up to)

However, even with the JIT enabled, plain LuaJIT still creates the desired output in my very simple test-case:
```
# rm luacov.stats.out
# luajit -e 'jit.on(); local luacov = require("luacov");' testthis.lua
testthis
# cat luacov.stats.out
5:testthis.lua
1 1 2 0 2 
```

So it is unclear what happens, but somehow Luvit prevents LuaCov from seeing the interpreter progress.

The same behaviour can be observed for [LuaTrace](https://github.com/geoffleyland/luatrace).

```
# luvit -e 'jit.off(); jit.flush(); local luatrace = require("luatrace"); luatrace.tron();' tests/run.lua
```

So apparently they both use the same path, which is blocked by Luvit.

[...]

Issue #896, opened 2016-05-16
Full path not displayed in traceback


SinisterRectus:
Currently, luvit's traceback shows strings for the source path. In the sample below, paths are appended with `string`, and in cases where the path is beyond a certain length, it is truncated by `...`.

```lua
Uncaught exception:
[string "bundle:deps/require.lua"]:291: [string "C:/Luvit/Discordia/deps/Discord
ia/classes/dis..."]:6: unexpected symbol near 'local'
stack traceback:
        [C]: in function 'assert'
        [string "bundle:deps/require.lua"]:291: in function 'require'
        [string "C:/Luvit/Discordia/deps/Discordia/classes/dis..."]:2: in function 'fn'
...
```

This is a result of the custom require module that luvit uses, where ever `loadstring` is called. From the Lua 5.2 manual:

> source: the source of the chunk that created the function. If source starts with a '@', it means that the function was defined in a file where the file name follows the '@'. If source starts with a '=', the remainder of its contents describe the source in a user-dependent manner. Otherwise, the function was defined in a string where source is that string.

So, in `require.lua`, if you append "@" to the path, this fixes half of the issue.

By changing
`local fn = assert(loadstring(data, path))`
to
`local fn = assert(loadstring(data, "@" .. path))`

, the trace becomes:

```lua
Uncaught exception:
[string "bundle:deps/require.lua"]:291: ...vit/Discordia/deps/Discordia/classes/
discord/message.lua:6: unexpected symbol near 'local'
stack traceback:
        [C]: in function 'assert'
        [string "bundle:deps/require.lua"]:291: in function 'require'
        C:/Luvit/Discordia/deps/Discordia/classes/discord/user.lua:2: in function 'fn'
...
```

This solves the issue of required file paths, but it leaves luvi bundled paths untouched. This would presumably be solved by applying a similar adjustment to the `loadstring` call in `luvibundle.lua` in the luvi source, but this is something that I have not been able to properly investigate.

(Shoutout to @SpaceManiac for helping discover a fix.)

creationix:
Thanks for the find.  If I understand correctly, this means that luvit-loader.lua based programs don't have this problem since they use native `dofile` to require files.

SinisterRectus:
That's probably true, since `load` and `loadstring` do not always load a file, but `dofile` is used exlusively for files.

SinisterRectus:
Sorry, keep accidentally closing.

creationix:
[...]

PR #922, opened 2016-08-16
Document how to use Luvit with LuaCov

urzds:
Fixes: #843


creationix:
Another solution would be to tweak require to follow the lua convention and include the `@` prefix when the source is a file.

urzds:
Trying to figure out how to implement your proposed change, I stumbled upon 74e947ab8ee2af87c46021acb135bd2bdbceb626 and 126264f85186c17e82f2d85c495132a9104d9b13 (related to #896), where you already fixed this. I just tested this and it works.

creationix:
Oh yeah, I forget that was in luvit itself.  For some reason I thought that was only in lit.

PR #923, opened 2016-08-18
[readline] Faster way the cover character to be replaced.

Miigon:
The results are the same(with multi characters supported).


deps/readline.lua
```diff
@@ -108,7 +108,7 @@ function Editor:insertAbove(line)
   end)
 end
 function Editor:insert(character)
-  local display = self.cover or character
+  local display = string.rep(self.cover, #character)
```
creationix: This is broken when `self.cover` is nil (the default setting).  We need some sort of conditional.

creationix:
Thanks!

PR #930, opened 2016-09-22
allow loading from 'modules' sub-dir as legacy version

kchenhal:
To be backward compatible with legacy version. 
We are using legacy version now, our libraries are located under 'modules' sub-dir.  Now we want to take advantage of the latest luvit. Can we add this fix so that our existing code will still work?

thanks
Kevin Chen

### DIFF
--- a/deps/boundary.lua
+++ b/deps/boundary.lua
@@ -1,0 +1,50 @@
+-- Copyright 2015 Boundary
+-- @brief convenience variables and functions for Lua scripts
+-- @file boundary.lua
+--
+--[[lit-meta
+  name = "luvit/boundary"
+  version = "2.0.0"
+  license = "Apache 2"
+  homepage = "https://github.com/luvit/luvit/blob/master/deps/core.lua"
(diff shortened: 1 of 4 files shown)


squeek502:
Note that there were many backwards-incompatible changes between Luvit 1.0 and Luvit 2.0.

kchenhal:
Yes, I understand. But this should not cause problem for the new version and will help us a lot. 

Some background:
We provide a framework that one of the functionality is to allow plugin(written as lua) to provide us data, we currently have a lot of the plugins. My goal is to make the existing plugins work with the upgrade.

thanks
Kevin

creationix:
@kchenhal You do know it's trivial to make your own custom luvit binary right?  This was one of the main goals of the luvit 2.0 rewrite.  The pre-built luvi binaries combined with the zip asset system and the lit tool make is very easy to tweak and pack a custom program forked from luvit.

kchenhal:
I have already closed the PR, I have more changes which some of them are not related to common use of luvit . I will keep those to my own repository just like you mentioned. 

thanks!!

PR #935, opened 2016-11-26
(#924) Support UTF-8 and fix readline

### LINKED ISSUE #924: Repl didn't do backspace correctly on multi-byte characters
@creationix 
This happened when I typed some multi-byte characters (such as chinese or japanese) and pressed backspace.

It will only delete last one byte of the character,not all of them(Unicode said they are 2 or 3 bytes)
so it will be...

```
> 中文
Length is 6
```

... and press backspace

```
> 中��
Length is 5
```

I must press backspace twice or three times to delete all bytes of a character.
The main issue is to get the actually length of each character.I think this can be fixed by adding a utf8 wrapper.
btw,this also cause the position of cursor to be wrong.

Miigon:
This fixed #924.
I've write a new module `ustring` to process UTF-8 strings.It has similar APIs to standard `string` module.
And with this new module,the issue was fixed.
**PS:I don't know how `lit` package management system works,but my changes may cause `readline` become depending on `ustring`.**

### DIFF
--- a/deps/readline.lua
+++ b/deps/readline.lua
@@ -18,6 +18,9 @@ limitations under the License.
 --[[lit-meta
   name = "luvit/readline"
   version = "2.1.0"
+  dependencies = {
+    "luvit/ustring@2.0.0",
+  }
   homepage = "https://github.com/luvit/luvit/blob/master/deps/readline.lua"
   description = "A readline interface for terminals in pure lua."
   tags = {"readline", "tty"}
--- a/deps/readline.lua
+++ b/deps/readline.lua
@@ -30,6 +30,7 @@ limitations under the License.
 local ustring = require "ustring"
 local emptyline = ustring.new()
 local sub = ustring.sub
+local chlen = ustring.chlen
 local gmatch = ustring.gmatch
 local remove = table.remove
 local insert = table.insert
@@ -397,7 +398,7 @@ local keyHandlers =
     self:getHistory(10)
   end},
   -- Printable characters
-  {{function(key, char) return char > 31 and key or nil end}, function(self, consumedKeys)
+  {{function(key, char) return char > 31 and key:sub(1,chlen(key:byte())) or nil end}, function(self, consumedKeys)
     self:insert(consumedKeys)
   end},
 }
--- a/deps/readline.lua
+++ b/deps/readline.lua
@@ -32,6 +32,7 @@ local emptyline = ustring.new()
 local sub = ustring.sub
 local chlen = ustring.chlen
 local gmatch = ustring.gmatch
+local find = ustring.find
 local remove = table.remove
 local insert = table.insert
 local concat = table.concat
@@ -219,7 +220,7 @@ local function findLeft(line, position, wordPattern)
   local s
   repeat
     local start = sub(line, 1, position - 1)
-    s = string.find(start, pattern)
+    s = find(start, pattern)
     if not s then
       position = position - 1
     end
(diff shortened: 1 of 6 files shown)


creationix:
Yes, this will need some packaging to work with lit properly, but I can add that.

creationix:
Does this failure look related to this PR?  It is a test for the readline module.
https://travis-ci.org/luvit/luvit/jobs/179034058#L3713-L3729

squeek502:
Yes, the tests will need to be updated, as the current test is testing for string equality. Here's the assert that's failing:

`assert(editor.line == "a")`

Would probably need to do `assert(tostring(editor.line) == "a")` or compare against a ustring.

creationix:
@squeek502 sounds good.  I like the tostring style.

squeek502:
Looks like the PR currently struggles with a mixture of control sequences and characters being added at the same time.

```
    editor:onKey('a\027[Db\027[Dc\027[Dd')
```

which should be interpreted as ('a', left arrow, 'b', left arrow, 'c', left arrow, 'd') is only resulting in 'd' being added to the line.

deps/readline.lua
```diff
@@ -396,7 +396,7 @@ local keyHandlers =
     self:getHistory(10)
   end},
   -- Printable characters
-  {{function(key, char) return char > 31 and key:sub(1,1) or nil end}, function(self, consumedKeys)
+  {{function(key, char) return char > 31 and key or nil end}, function(self, consumedKeys)
```
squeek502: This should probably be `sub(key, 1, 1)` instead of removing the `sub` entirely. Removing the `sub` causes the entire line to get treated as printable whenever the first character is a printable character (e.g. 'a\027[D' should only consume 'a' and then interpret the rest as a left arrow key press, but without the sub, it will print the entire sequence instead).

deps/ustring.lua
```diff
@@ -0,0 +1,188 @@
...
+end
+
+local gsub = string.gsub
+local find = string.find
+local format = string.format
+local gmatch = string.gmatch
+local match = string.match
+local lower = string.lower
+local upper = string.upper
+ustring.len = rawlen
```
squeek502: `rawlen` is never defined

deps/ustring.lua
```diff
@@ -0,0 +1,188 @@
...
+
+function ustring.lower(ustr)
+    local u = ustring.copy(ustr)
+    for i = 1,#u do
+        u = lower(u)
+    end
+    return u
+end
+
+function ustring.upper()
```
squeek502: Missing the `ustr` param here, `ustr` is undefined in the line below

deps/ustring.lua
Miigon: `rawlen`is lua's builtin function.
https://www.lua.org/manual/5.2/manual.html#pdf-rawlen

deps/readline.lua
Miigon: `key:sub(1,1)`suppose that the printable character is 1 byte length,but unfortunately,Lot's of unicode character isn't 1-byte-wide.This sub only gets the first byte of the character.

Maybe we need a function `chlen()` to get the length of a unicode character and the code will be like `key:sub(1,ustring.chlen(key))`

Miigon:
Appveyor still failed,but I don't have a Windows environment to debug with.Sorry.

squeek502:
Build is failing because `ustring` has not been added to [`package.lua`](https://github.com/luvit/luvit/blob/master/package.lua), so it's not being included when building luvit:

```
[string "bundle:deps/require.lua"]:278: No such module 'ustring' in 'bundle:/deps/readline.lua'
```

deps/readline.lua
squeek502: Since `sub` was defined as `local sub = ustring.sub`, I assumed that it would account for the unicode characters. Does `ustring.sub` not work when called on a regular string?

Miigon:
@squeek502 So the only thing I need to do is adding it to `package.lua`,right?

creationix:
@Miigon I believe so.  Try that.

PR #936, opened 2016-11-29
Added write methods to Buffer

SinisterRectus:
Along with some minor formatting fixes, I've added write methods to complement the existing read ones. Some things to note:

- The parameters are `(offset, value)` instead of `(value, offset)`. The latter is used in Node, but I think the former better matches Lua and Luvit.
- Odd behavior is seen when writing 32-bit ints, presumably due to how [LuaJIT handles them](http://bitop.luajit.org/semantics.html). To compensate, `bit.rshift(value, 0)` is used when writing the LSB.
- There are other feature's of [Node's Buffer class](https://nodejs.org/api/buffer.html) that can be implemented, but I think we can do without them for now.


creationix:
> The parameters are (offset, value) instead of (value, offset). The latter is used in Node, but I think the former better matches Lua and Luvit.

I agree that `(offset, value)` is better, but since this API exists in node I worry changing the behavior but keeping the same name will confuse people. 

What do others think?

SinisterRectus:
I also made that decision with `setTimeout(delay, callback)` vs `setTimeout(callback, delay)` in mind.

creationix:
Good point.

creationix:
I wonder if this needs a rebase since it's conflicting?

SinisterRectus:
Couldn't quite get the rebase working, so I went with a merge. Hope that's okay.

PR #942, opened 2016-12-31
fix fs.ReadStream doesn't return anything

### LINKED ISSUE #940: Cannot arbitrarily read data from a stream
The following does not work as expected.

```lua
local fs = require('fs')
local stream = require('stream')

local file = io.open('test.txt', 'w')
file:write('abcdef')
file:close()

file = fs.createReadStream('test.txt')
p(file:read(6)) -- nil

file = fs.createReadStream('test.txt')
local sink = stream.Writable:new()
function sink:write(chunk) p(chunk) end
file:pipe(sink) -- 'abcdef'
```

When I attempt to read the stream directly, I get `nil`. Piping the stream to a different one works, though. I've tried all combination of `pause`, `resume` and `on:('readable', ...)` that I can think of, but can't seem to read from streams with anything other than `on:('data', ...)` . Am I doing something wrong?

Miigon:
this fixes #940

### DIFF
--- a/deps/fs.lua
+++ b/deps/fs.lua
@@ -658,18 +658,17 @@ function fs.ReadStream:_read(n)
     end
   end
 
-  fs.read(self.fd, to_read, self.offset, function(err, bytes)
-    if err then return self:destroy(err) end
-    if #bytes > 0 then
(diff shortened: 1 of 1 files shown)


creationix:
Thanks!

PR #944, opened 2017-01-23
fix stream closing

rphillips:
The exit callback from the spawn gets triggered before the data from the pipes is read and flushed. That exit callback calls `shutdown` on those pending pipes before they are done.

This change causes the pipe streams to automatically shutdown when they get closed by the stream. The issue only manifests itself on a busy server, and when multiple spawns are called on the same tick, which is why we have not seen it before.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+# Changes in 2.12.1
+
+ - childprocess: Fix spawn to cleanup correctly (Ryan Phillips)
+
 # Changes in 2.12.0
 
  - childprocess: Fix `childprogress.exec` return. (Cyril Hou)
--- a/deps/childprocess.lua
+++ b/deps/childprocess.lua
@@ -64,21 +64,9 @@ function Process:close(err)
 end
 
 function Process:destroy(err)
-  self:_cleanup(err)
   if err then
     timer.setImmediate(function() self:emit('error', err) end)
   end
-end
-
-function Process:_cleanup(err)
-  timer.setImmediate(function()
-    if self.stdout then
-      self.stdout:_end(err) -- flush
-      self.stdout:destroy(err) -- flush
-    end
-    if self.stderr then self.stderr:destroy(err) end
-    if self.stdin then self.stdin:destroy(err) end
-  end)
 end
 
 local function spawn(command, args, options)
@@ -155,6 +143,10 @@ local function spawn(command, args, options)
   em = Process:new(stdin, stdout, stderr)
   em:setHandle(handle)
   em:setPid(pid)
+
+  if stdout then stdout:resume() end
+  if stderr then stderr:resume() end
+  if stdin then stdin:resume() end
 
   if not em.handle then
     timer.setImmediate(function()
--- a/deps/childprocess.lua
+++ b/deps/childprocess.lua
@@ -144,18 +144,21 @@ local function spawn(command, args, options)
   em:setHandle(handle)
   em:setPid(pid)
 
-  if stdout then stdout:resume() end
-  if stderr then stderr:resume() end
-  if stdin then stdin:resume() end
-
   if not em.handle then
     timer.setImmediate(function()
       em.exitCode = -127
       em:emit('exit', em.exitCode)
-      em:destroy(Error:new(pid))
+      em:emit('error', Error:new(pid))
+      if em.stdout then em.stdout:emit('error', Error:new(pid)) end
+      if em.stderr then em.stderr:emit('error', Error:new(pid)) end
+      if em.stdin then em.stdin:emit('error', Error:new(pid)) end
       maybeClose()
     end)
   end
+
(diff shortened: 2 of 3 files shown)


rjemanuele:
@jjbuchan It looks like @rphillips code is removing the old process cleanup code in favor of resuming the stdio operations when the process exits.  Then the stdio streams will be gc'd once they are no longer needed by the process once it is destroyed.

rphillips:
commented in the description.

jjbuchan:
Thanks.  The theory makes sense to me and it seems worth a try.

rphillips:
Test program

```lua
local uv = require('uv')
local net = require('net')
local childprocess = require('childprocess')
local timer = require('timer')
local LineEmitter = require('line-emitter').LineEmitter
local async = require('async')

local function run(callback)
	local stdoutLineEmitter = LineEmitter:new()
	local stdio = {
		nil,
		net.Socket:new({ handle = uv.new_pipe(false) }),
		nil
	}
	local child = childprocess.spawn('./test_plugin.sh', {'./test_plugin.sh'}, { stdio = stdio })
	local count = 0
	local lines = {}

	stdoutLineEmitter:on('data', function(line)
		count = count + 1
		table.insert(lines, line)
	end)

	child.stdout:pipe(stdoutLineEmitter)

	local function waitForIO(callback)
		stdoutLineEmitter:once('end', function()
		  callback()
        	end)
        end

        local code = -1

        local function waitForExit(callback)
		child:once('exit', function(_code)
			code = _code
			callback()
		end)
        end

        async.parallel({ waitForIO, waitForExit }, function()
        		if code ~= 0 then
        			p('code ~= 0', code)
        			return
        		end
			p(lines)
			if count ~= 2 then
				p(child)
			end
			assert(code == 0)
			assert(count == 2)
			callback()
        end)
end

local function go()
	local t = timer.setTimeout(10, run, function()
		timer.setTimeout(10, go)
	end)
end

timer.setInterval(20, run, function() end)
timer.setInterval(20, run, function() end)
timer.setInterval(20, run, function() end)
timer.setInterval(20, run, function() end)
timer.setInterval(20, run, function() end)
timer.setInterval(20, run, function() end)
```

creationix:
node streams in luvit sure are strange.  As long as `:resume()` has the intended effect, then +1

joerg-krause:
This commit changes the behavior of `exec()`.

Test program:

```lua
local exec = require("childprocess").exec
local child = exec("uname", function (err, stdout, stderr)
  p(err, stdout, stderr)
end)
child:on("close", function (code, signal) p("close", code, signal) end)
child:on("error", function (err) p("error", err) end)
child:on("exit", function (code, signal) p("exit", code, signal) end)
child.stdout:on("data", function (data) p(data) end)
```

Output with Luvit 2.14.2:
```
'data:'	'Linux\n'
'exit:'	0	0
'callback:'	nil	'Linux\n'	''
'close:'	0	0
```

Output with Luvit 2.15.0:
```
'data:'	'Linux\n'
'exit:'	0	0
```

joerg-krause:
Looks like this issue depends on the version of Luvi used. Version 2.7.6 is working fine, whereas 2.9.1 does not. Reverting Luvits commit ab8bcaaf1f813914410bca53c4d47e68532f12cf fixes the issue with Luvi 2.9.1.

Doing a git-bisect for Luvi shows that the issue was introduced in Luvis commit https://github.com/luvit/luvi/commit/8841d3c42f6b95d7da437cd8abff5d663723147a.

squeek502:
@joerg-krause as a temporary workaround for the specific example code you posted, you could use [`uv.os_uname()` added to Libuv/Luv in 1.25.0](https://github.com/luvit/luv/blob/master/src/misc.c#L496-L517) ([Libuv docs](http://docs.libuv.org/en/v1.x/misc.html#c.uv_os_uname))

joerg-krause:
Thanks, but actually this is only an example. Actually, I am running a different program then uname.

joerg-krause:
The precise libuv commit is https://github.com/libuv/libuv/commit/c409b3fcff51ea270724e4f43866764e65666095, which introduces failing the test program.

rphillips:
@joerg-krause looks like something like this might fix luvit: https://github.com/libuv/libuv/pull/1655#issuecomment-358673030

joerg-krause:
I am happy to test any solutions you propose. I am out of my depth here...

joerg-krause:
@rphillips Do you have a test case to check if commit ab8bcaaf1f813914410bca53c4d47e68532f12cf is really necessary? Maybe it is safe to revert this commit?

joerg-krause:
Has anyone an idea how to proceed?

joerg-krause:
https://github.com/luvit/luvit/pull/944#issuecomment-469864902 is fixed by   9d14161ee72a7d85aa1c8bdb3435427514738786.

PR #945, opened 2017-01-31
dns: Add the ability to handle numeric addresses

n1tehawk:
When users specify a numeric address (IPv4 or IPv6) for the host, the `dns` module will currently construct a DNS query from it and send/forward that to the 'upstream' servers. This is likely to simply fail with a `3: name error` in most cases (depending on the DNS server).

This pull request enables `dns` to detect and properly handle such "names", by resolving them into a single-address 'reply' internally.

Regards, NiteHawk


creationix:
+1

creationix:
@rphillips looks good to me, I'll give you a chance to look it over before merging.

rphillips:
+1 lgtm as well

PR #946, opened 2017-02-01
dns: Fix IPv6 address matching

n1tehawk:
The previous pattern for IPv6 address detection was overly simply.
This new filter requires at least one ':' to be present within the
address, and it also allows for dots in the notation of IPv4-mapped
IPv6 (e.g. ::ffff:192.0.2.128).

See https://en.wikipedia.org/wiki/IPv6_address#Representation


rphillips:
nice. +1

creationix:
Thanks for the quick followup.

PR #952, opened 2017-03-04
Remove most optional deps since they are also in userspace

creationix:
In an effort to get closer to the proposed luvit 3.x that's a minimal core, this PR removes most the optional standard libraries that are also available via lit.


Haider982:
any updates?

PR #960, opened 2017-04-10
Add http status code of 429

yaa110:


creationix:
Looks good.
Can you also double check to see if we have all the new codes from https://tools.ietf.org/html/rfc6585

yaa110:
A new PR #974 was opened. So, close this one.

PR #969, opened 2017-04-18
Unbreak stream example

joerg-krause:
Fixes:
```
luvit test.lua 
Uncaught exception:
/root/stream/libs/file.lua:67: attempt to index local 'iStream' (a nil value)
```

### DIFF
--- a/examples/stream/libs/file.lua
+++ b/examples/stream/libs/file.lua
@@ -1,7 +1,9 @@
 -- This is a partial port of the built-in fs module as continuable format
 
 local native = require('uv')
-local iStream = require('stream').iStream
+local Emitter = require('core').Emitter
+
+local iStream = Emitter:extend()
 
 local fs = {}
 


examples/stream/libs/file.lua
```diff
@@ -1,7 +1,11 @@
 -- This is a partial port of the built-in fs module as continuable format
 
 local native = require('uv')
-local iStream = require('stream').iStream
+local Emitter = require('core').Emitter
+local iStream = Emitter:extend()
+
+local core = {}
```
creationix: Why do you create a `core` table here?

joerg-krause: Because of the following line `core.iStream = iStream`.

joerg-krause:
I've removed the core table and updated the commit.

creationix:
Looks great, thanks!

PR #970, opened 2017-04-19
Update require.lua

zhaozg:
* Fix require to look in bundle *after* filesystem

### DIFF
--- a/deps/require.lua
+++ b/deps/require.lua
@@ -190,6 +190,12 @@ local function moduleRequire(base, name)
       base = pathJoin(base, "..")
     end
   end
+  -- Look in bundle first to avoid conflicts when developing luvi apps.
+  if not base:match("^bundle:/*") then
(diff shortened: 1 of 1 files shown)


creationix:
What is this trying to do?

The logic I want (I think) is to look in the bundle as a fallback if the dependency isn't found in the filesystem. This is for require being called in a module outside the bundle.  A module in the bundle should only look for dependencies in the bundle.

zhaozg:
Right.
The PR can solve the problem as your descible. 

A few luvit apps, they share a luvit. In development stage, projects layout is
```
|--luvit
|--appA
|--appB
|--cppC
|--luvi.exe
|--luvit.bat
```
Finally all projects lookup module from luvit/deps and [app]/libs.

rphillips:
@zhaozg is this still needed? Travis failed the build

PR #977, opened 2017-05-03
Fix bundle requires on Windows

squeek502:
Bundle requires were only being loaded when reaching / as root, but something like C: is the root on Windows, so the bundle was never being looked in (because it was assuming something like C: was the bundle root rather than the filesystem root).

(see comments in https://github.com/luvit/luvit/pull/973)


SinisterRectus:
Fixes the issue. Thanks!

Issue #950, opened 2017-02-26
Reading from /dev/input/event0 does not work anymore


joerg-krause:
Commit https://github.com/luvit/luvit/commit/edf48cf3093a574805595e612680c006a6cdf741 was added by @Miigon to fix https://github.com/luvit/luvit/issues/940, but unfortunately breaks reading from `/dev/input/event0`, which was working fine before.

The following code snippet does not work anymore:

``` Lua
local fs = require("fs")
local Emitter = require("core").Emitter
local KeyEventListener = Emitter:extend() -- class table

function KeyEventListener:initialize(id)
  self.id = id
  self:wrap("onOpen")
  self:wrap("onRead")
  fs.open("/dev/input/event" .. id, "r", "0644", self.onOpen)
end

function KeyEventListener:onOpen(fd)
  self.fd = fd
  self:startRead()
end

function KeyEventListener:startRead()
  fs.read(self.fd, ffi.sizeof("struct input_event"), nil, self.onRead)
end

function KeyEventListener:onRead(data)
  p(data)
  if self.fd then self:startRead() end
end

function KeyEventListener:close(callback)
  local fd = self.fd
  self.fd = nil
  fs.close(fd, callback)
end

return {
  KeyEventListener = KeyEventListener
}
```

The callback function `onRead()` passed to `fs.read()` is never called on an input event. Reverting this commit helps.

creationix:
:(

Miigon:
Since I just replaced `fs.read` with `fs.readSync`,maybe these sockets don't support `fs.readSync`?

creationix:
@Miigon I think the problem is you turned a non-blocking function into a blocking function and so it no longer invokes the callback.

I don't know this API very well, but as I commented in your commit, I think it's supposed to be non-blocking.

joerg-krause:
Fixed in version 2.14.1.

PR #980, opened 2017-05-15
Release 2.14.1

creationix:
# Changes in 2.14.1

 - dns: Allow ip addresses in dns resolve to resolve internally. [NiteHawk]
 - http-codec: Add HTTP status codes of RFC 6585 [Navid]
 - require: Fix bundle requires on Windows [Squeek]
 - Support punctuation in http credentials [Charles Farquhar]
 - Update lit to 3.5.4 [Tim Caswell]

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,7 @@
  - require: Fix bundle requires on Windows [Squeek]
  - Support punctuation in http credentials [Charles Farquhar]
  - Update lit to 3.5.4 [Tim Caswell]
+ - Revert fs change that broke readstreams [Tim Caswell]
 
 # Changes in 2.12.1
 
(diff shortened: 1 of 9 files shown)


joerg-krause:
Any intention to resolve https://github.com/luvit/luvit/issues/950? The agreement was to revert commit https://github.com/luvit/luvit/commit/edf48cf3093a574805595e612680c006a6cdf741.

creationix:
@joerg-krause Good catch, anything else?

SinisterRectus:
Looks good as far as I can tell.

creationix:
Hmm, the repl is broken for me (try ALT + Arrows or Control + Arrows on gnome terminal)  It should be jump word or home/end just like in bash, but instead if crashes.  I suspect it's related to the new ustring code.  I'm going to publish this for now, but we should fix that soon.

joerg-krause:
@creationix Thanks! Nothing else from my point of view.

PR #992, opened 2017-06-11
Miscellaneous fixes and optimizations

SinisterRectus:
- `byte` was undefined in one place in `url.lua`. To stay consistent with the rest of the file, I changed to `string.byte`.
- I don't know how many people rely on `string.levenshtein` or even know that it's there, but I've optimized it in a few places.
- Many of the changes are me nit-picking some concatenations. In `repl.lua`, I moved the newlines to a new write call and in `pretty-print.lua`, I moved the open/close characters to new write calls. This is especially good for writing large strings; I was able to cut garbage creation by about 50% in some cases.
- Also in `pretty-print.lua`, I changed a few `match` calls to `find` (functional equivalency without returning a new string) and changed the final behavior of `prettyPrint` to just write straight to stdout. I don't see any point in passing the args through `print` after they've already been concatenated.
- Finally, in `pretty-print.lua` again, I removed a duplicate `strip` definition.

### DIFF
--- a/deps/helpful.lua
+++ b/deps/helpful.lua
@@ -26,34 +26,33 @@ limitations under the License.
 ]]
 
 function string.levenshtein(str1, str2)
-  local len1 = string.len(str1)
-  local len2 = string.len(str2)
-  local matrix = {}
-  local cost
-  -- quick cut-offs to save time
-  if (len1 == 0) then
-    return len2
-  elseif (len2 == 0) then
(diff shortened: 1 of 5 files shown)


SinisterRectus:
My http-codec change broke coro-http, presumably because it relies on the table count being 1 at the start. Rather than try to fix it and break it more, I just reverted it.

PR #993, opened 2017-06-14
Changes servername from host to hostname

SinisterRectus:
`hostname` with no port, should be used instead of `host`.

Try GET https://luvit.io:443/ with hostname vs host.

https://nodejs.org/api/tls.html#tls_server_addcontext_hostname_context


SinisterRectus:
Looks like the TLS tests and the HTTP(S) modules look for host, not hostname. Only the url module makes a distinction between the two. I guess without a lot of re-writing, we'll have to keep this one as host.

rphillips:
yeah, makes sense

PR #1002, opened 2017-07-31
Adds missing __newindex for require global

SinisterRectus:
Not sure if this is considered a breaking change or a fix.

Currently, you cannot set globals without explicitly indexing `_G`
```lua
randomGlobal = 'module global value'
_G.otherGlobal = 'actual global value'
```
By setting the `__newindex` of the module to `_G`, both options become equivalent.

### DIFF
--- a/deps/require.lua
+++ b/deps/require.lua
@@ -306,7 +306,7 @@ function Module:require(name)
         return module:require(...)
       end
     }
-    setfenv(fn, setmetatable(global, { __index = _G }))
+    setfenv(fn, setmetatable(global, { __index = _G, __newindex = _G }))
     local ret = fn()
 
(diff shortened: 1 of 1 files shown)


creationix:
It was kinda a feature to avoid accidental global leakage.  I can see how people who already know lua get bitten by it though.

SinisterRectus:
I added this because I've seen people get tripped up by this quirk, but once they learn that they need to directly assign to `_G`, then it is no longer an issue (except when someone says "that's dumb, why do I need to do that").

I'll leave the decision up to you. I don't mind keeping things as they are.

creationix:
I'm worried that changing it will break things.  Let's leave it alone for now.

PR #1007, opened 2017-09-06
Flush streams on exit (fixes https://github.com/luvit/lit/issues/222)

creationix:


SinisterRectus:
I was going to say that you could move the `close` callback outside of the `walk` loop and pass `handle` to it instead of creating a new callback with each pass.

creationix:
I could, but then I'd have to create a closure for each stream so that I had the handle to pass to the common close.

I don't think it's a hot path either way.

SinisterRectus:
Or maybe not, if you have to pass the closure to `shutdown`.

joerg-krause:
This PR breaks piping the output of Luvit to a file. See #1083.

Issue #1013, opened 2017-10-28
Get CPU and RAM usage


Ale32bit:
Is there a way to get the CPU and RAM usage?

squeek502:
You can using the [`luv` bindings](https://github.com/luvit/luv/blob/96591671bc468b842b11301e9b374b13b7b5c768/src/luv.c#L225-L262).

```
local uv = require('uv')

local cpuinfo = uv.cpu_info()
p(cpuinfo)

local totalMem = uv.get_total_memory()
local freeMem = uv.get_free_memory()
local usedMem = totalMem - freeMem
print(string.format("%d used (out of %d)", usedMem, totalMem))
```

`uv.cpu_info()` will output [the same thing as node's `os.cpus()`](https://nodejs.org/api/os.html#os_os_cpus), and can be [used in similar ways to get % usage, etc](https://stackoverflow.com/questions/36816181/get-view-memory-cpu-usage-via-nodejs) (wouldn't be too hard to [port this to luvit](https://github.com/oscmejia/os-utils/blob/master/lib/osutils.js#L165-L213)).

`uv.get_total_memory` and `uv.get_free_memory` return the number of bytes total/free (the same as node's `os.totalmem`/`os.freemem`).

PR #1014, opened 2017-10-28
Add process:cpuUsage() and process:memoryUsage() functions

squeek502:
Based on the node functions of the same name:
 - Luvit's `cpuUsage` works identically to node's (https://nodejs.org/api/process.html#process_process_cpuusage_previousvalue)
 - Luvit's `memoryUsage` works identically to node's (https://nodejs.org/api/process.html#process_process_memoryusage), but does not include the `heapTotal` or `external` keys.

(related: #1013)


SinisterRectus:
Shouldn't `memoryUsage` return `{rss = resident_set_memory}` instead of just `resident_set_memory`?

SinisterRectus:
Node uses v8 magic to get heap statistics

```cpp
void Isolate::GetHeapStatistics(HeapStatistics* heap_statistics) {
  i::Isolate* isolate = reinterpret_cast<i::Isolate*>(this);
  i::Heap* heap = isolate->heap();
  heap_statistics->total_heap_size_ = heap->CommittedMemory();
  heap_statistics->total_heap_size_executable_ =
      heap->CommittedMemoryExecutable();
  heap_statistics->total_physical_size_ = heap->CommittedPhysicalMemory();
  heap_statistics->total_available_size_ = heap->Available();
  heap_statistics->used_heap_size_ = heap->SizeOfObjects();
  heap_statistics->heap_size_limit_ = heap->MaxReserved();
  heap_statistics->malloced_memory_ =
      isolate->allocator()->GetCurrentMemoryUsage();
  heap_statistics->peak_malloced_memory_ =
      isolate->allocator()->GetMaxMemoryUsage();
  heap_statistics->does_zap_garbage_ = heap->ShouldZapGarbage();
}
```
```cpp
static void MemoryUsage(const FunctionCallbackInfo<Value>& args) {
  Environment* env = Environment::GetCurrent(args);

  size_t rss;
  int err = uv_resident_set_memory(&rss);
  if (err) {
    return env->ThrowUVException(err, "uv_resident_set_memory");
  }

  Isolate* isolate = env->isolate();
  // V8 memory usage
  HeapStatistics v8_heap_stats;
  isolate->GetHeapStatistics(&v8_heap_stats);

  // Get the double array pointer from the Float64Array argument.
  CHECK(args[0]->IsFloat64Array());
  Local<Float64Array> array = args[0].As<Float64Array>();
  CHECK_EQ(array->Length(), 4);
  Local<ArrayBuffer> ab = array->Buffer();
  double* fields = static_cast<double*>(ab->GetContents().Data());

  fields[0] = rss;
  fields[1] = v8_heap_stats.total_heap_size();
  fields[2] = v8_heap_stats.used_heap_size();
  fields[3] = isolate->AdjustAmountOfExternalAllocatedMemory(0);
}
```

deps/process.lua
```diff
@@ -124,6 +124,31 @@ local function exit(self, code)
   process.stderr:_end()
 end
 
+-- Returns the resident set size of the current process in bytes
+local function memoryUsage(self)
+  return uv.resident_set_memory()
```
SinisterRectus: Shouldn't this return {rss = resident_set_memory} instead of just resident_set_memory?

squeek502: I simplified that return value since we aren't returning the other heap stats, but it could definitely still be a table for compatibility/future extensibility.

squeek502:
@SinisterRectus updated memoryUsage() to return a table, and added a `heapUsed` key for what I think is the Luvit equivalent of node's `heapUsed`.

creationix:
+1

SinisterRectus:
Looks good. I don't know if collectgarbage count is equivalent to heap memory, but it's probably the best available number.

creationix:
I think it's close enough.  Pretty much all garbage collectable memory is heap memory.  Do you know exactly what the node version counts?

SinisterRectus:
Not sure. This is as far as I am able to dissect it without diving too deep into how V8 manages memory.

```cpp
heap_statistics->used_heap_size_ = heap->SizeOfObjects();
heap_statistics->total_heap_size_ = heap->CommittedMemory();
```
```cpp
size_t Heap::SizeOfObjects() {
  size_t total = 0;
  AllSpaces spaces(this);
  for (Space* space = spaces.next(); space != NULL; space = spaces.next()) {
    total += space->SizeOfObjects();
  }
  return total;
}

size_t Heap::CommittedMemory() {
  if (!HasBeenSetUp()) return 0;

  return new_space_->CommittedMemory() + CommittedOldGenerationMemory();
}
```

creationix:
My understanding is the luajit one simply counts all GC managed memory.  From the little I know of V8 internals, I could see it being more complicated there.

PR #1028, opened 2018-05-25
remove keeplive on listen socket

zhaozg:
SO_KEEPALIVE is a TCP concept, but can apply to listen socket.
On Linux/Windows, it will be ignored, but  on macOS, then listen socket will be closed(not trigger any event) after timeout. That will make my server offline. 

This RP will solve this problem.

Cc @creationix


zhaozg:
Ci is red, but not related this PR.

creationix:
:+1:

PR #1041, opened 2018-09-15
update default TLS from TLSv1 to TLSv1_2

zhaozg:


zhaozg:
I will check on my host

PR #1046, opened 2018-09-27
Create hard links on Windows

shawwn:
This is a better version of #1045. These hard links allow `npx luvit` to work on Windows.


rphillips:
Thanks!

PR #1047, opened 2018-10-13
Match only past last . in file extensions

truemedian:
allows files such as file.test.lua and file.something.else.lua to both be required without an unknown type error


rphillips:
thanks!

PR #1048, opened 2018-10-30
Add uncaughtException handler

Python1320:
Similar to https://nodejs.org/api/process.html#process_event_uncaughtexception and the crash stack can also be observed.

I will need to add a test case and see if there is a better way to implement this.

### DIFF
--- a/deps/process.lua
+++ b/deps/process.lua
@@ -87,7 +87,7 @@ end
 local signalWraps = {}
 
 local function on(self, _type, listener)
-  if _type == "error" or _type == "exit" then
+  if _type == "error" or _type == "uncaughtException" or _type == "exit" then
     Emitter.on(self, _type, listener)
   else
     if not signalWraps[_type] then
@@ -212,6 +212,7 @@ local function globalProcess()
   process.stdout = UvStreamWritable:new(pp.stdout)
   process.stderr = UvStreamWritable:new(pp.stderr)
   hooks:on('process.exit', utils.bind(process.emit, process, 'exit'))
+  hooks:on('process.uncaughtException', utils.bind(process.emit, process, 'uncaughtException'))
   return process
 end
 
--- a/init.lua
+++ b/init.lua
@@ -50,7 +50,10 @@ return function (main, ...)
 
     -- Start the event loop
     uv.run()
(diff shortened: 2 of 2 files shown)


SinisterRectus:
Is there a reason why you call the hook in the error handler instead of after xpcall returns:

```lua
  if success then
    -- Allow actions to run at process exit.
    require('hooks'):emit('process.exit')
    uv.run()
  else
    _G.process.exitCode = -1
    require('hooks'):emit('process.uncaughtException', err)
    require('pretty-print').stderr:write("Uncaught exception:\n" .. err .. "\n")
end
```

Python1320:
It is so that we can use debug library to inspect the stack trace in a greater depth than what debug.traceback() can give, such as https://github.com/ignacio/StackTracePlus (not tested).

SinisterRectus:
I wonder if it makes sense to move the exit event to the xpcall handler then.

It's useful to have an uncaught exception handler, but it's also one of those things can trip up people who use it incorrectly. I'd wait for @creationix to weight in.

creationix:
I'm sorry, I've been away from Lua for the last few months and can't really offer meaningful technical feedback at the moment.

Also I'm not personally invested in the legacy node.js compat layer in the `luvit/luvit` project.  I'm much more interested in the newer APIs used in the luvi / lit / coro-* stacks.

I'll defer on this one to what you all think.

SinisterRectus:
I agree that avoiding a node.js API is best, but the current `luvit/luvit` package doesn't make this easy. Until we move the entire luvit project away from being a node clone (luvit 3.0 anyone?), I think this is an okay addition. The global process module will always be in 2.x.

Mehgugs:
I dont see any problem with adding this, however I don't think it's very useful? It would probably be better to just improve the traceback luvit produces by default directly if that is the only use case.

Python1320:
The root reason for this pull request is actually https://sentry.io for which this hook would be quite useful.

SinisterRectus:
Any objections before I pull this?

rphillips:
lgtm

SinisterRectus:
Totally forgot that I volunteered to pull this.

PR #1052, opened 2019-02-24
 Fix http-codec failing occasionally during chunked decoding 

squeek502:
It is not guaranteed that the data we have received so far will include both the chunk-size and its terminator, so failing in that case is wrong. Instead, we should wait until we get more data and check again. This commit makes it so that decoding only fails if we still don't find a terminator with a chunk-size field of >= 32 hex digits, which should never happen under normal circumstances.

Detailed example of when this can occur:

- Data received so far: `'4\r\nTest\r\n13\r'`
- First chunk decoded: len=`'4'`, data=`'Test'`
- Data remaining to decode: `'13\r'`
- Attempt to decode next chunk: len=`'1'`, term=`'3\r'`; term ~= `'\r\n'` (without this commit, the code would error here with a failed assert)
- New data received: `'\nFixed in this case\r\n'`
- Data remaining to decode: `'13\r\nFixed in this case\r\n'`
- Re-attempt to decode chunk: len=`'13'`, data=`'Fixed in this case'`

### DIFF
--- a/deps/http-codec.lua
+++ b/deps/http-codec.lua
@@ -251,7 +251,11 @@ local function decoder()
     local len, term
     len, term = match(chunk, "^(%x+)(..)")
     if not len then return end
-    assert(term == "\r\n")
+    if term ~= "\r\n" then
+      if #chunk < 34 then return end
+      -- But protect against evil clients by refusing chunk-sizes longer than 32 hex digits.
+      error("chunk-size field too large")
+    end
     local length = tonumber(len, 16)
     if #chunk < length + 4 + #len then return end
     if length == 0 then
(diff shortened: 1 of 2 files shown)


deps/http-codec.lua
```diff
@@ -251,7 +251,11 @@ local function decoder()
     local len, term
     len, term = match(chunk, "^(%x+)(..)")
     if not len then return end
-    assert(term == "\r\n")
+    if term ~= "\r\n" then
+      if #chunk < 34 then return end
```
alkino: 34 ?

squeek502: 32 hex digits + '\r\n'. It's an arbitrary limit, wasn't sure what to go with.

rphillips: would 18 make sense? `FFFFFFFFFFFFFFFF` (16 characters + 2 for \r\n) is the max int for len.

squeek502: Yeah, that would probably make more sense.

Issue #1055, opened 2019-03-05
uv.spawn() might fail


joerg-krause:
[uv.spawn()](https://github.com/luvit/luvit/blob/master/deps/childprocess.lua#L133) might fail, e.g. because of `ENOMEM: not enough memory`. In this case the `handle` is `nil` and the `pid` has the value of the error message.

`uv.spawn()` not returning a valid `handle` should be handled as an error and reported to the user.

squeek502:
Similar issue from when this affected `coro-spawn`: https://github.com/luvit/lit/pull/220

joerg-krause:
Closed, as the errors from uv.spawn() are actually handled. Not the way I thought it should be, but after studying the sources, e.g. looking at [test-process.lua](https://github.com/luvit/luvit/blob/master/tests/test-process.lua), I understand how to get the error messages.

PR #1056, opened 2019-03-06
childprocess: handle errors from uv.spawn

joerg-krause:
`uv.spawn()` might fail and return a `nil` handle and an error message
instead of a `pid`.

`spawn`, `exec`, and `execFile` will now return a `nil` handle and the error
message when `uv.spawn()` fails

Related to: https://github.com/luvit/lit/issues/219

Fixes: #1055

### DIFF
--- a/deps/childprocess.lua
+++ b/deps/childprocess.lua
@@ -123,11 +123,13 @@ local function spawn(command, args, options)
   end
 
   function onExit(code, signal)
-    em.exitCode = code
-    em.signal = signal
-    em:emit('exit', code, signal)
-    maybeClose()
-    em:close()
+    if em then
+      em.exitCode = code
+      em.signal = signal
+      em:emit('exit', code, signal)
+      maybeClose()
+      em:close()
+    end
   end
 
   handle, pid = uv.spawn(command, {
@@ -139,6 +141,10 @@ local function spawn(command, args, options)
     uid = options.uid,
     gid = options.gid
   }, onExit)
+
+  if not handle then
+    return nil, pid -- pid contains the error message
+  end
 
   em = Process:new(stdin, stdout, stderr)
   em:setHandle(handle)
@@ -200,7 +206,11 @@ local function _exec(file, args, options, callback)
     if not options[k] then options[k] = v end
   end)
 
-  local child = spawn(file, args, options)
+  local child, err = spawn(file, args, options)
+
+  if err then
+    return nil, err
+  end
 
   local stdout, stderr = {}, {}
   local exited = false


squeek502:
Looking at the code/tests, was this not handled already?

```lua
  handle, pid = uv.spawn(command, ...)

  em = Process:new(stdin, stdout, stderr)
  em:setHandle(handle)
  em:setPid(pid)

  if not em.handle then
    timer.setImmediate(function()
      em.exitCode = -127
      em:emit('exit', em.exitCode)
      em:emit('error', Error:new(pid))
      if em.stdout then em.stdout:emit('error', Error:new(pid)) end
      if em.stderr then em.stderr:emit('error', Error:new(pid)) end
      if em.stdin then em.stdin:emit('error', Error:new(pid)) end
      maybeClose()
    end)
end
```

```lua
  test('invalid command', function(expect)
    local child, onError

    -- disable on windows, bug in libuv
    if los.type() == 'win32' then return end

    function onError(err)
      assert(err)
    end

    child = spawn('skfjsldkfjskdfjdsklfj')
    child:on('error', expect(onError))
    child.stdout:on('error', expect(onError))
    child.stderr:on('error', expect(onError))
  end)

  test('invalid command verify exit callback', function(expect)
    local child, onExit, onClose

    -- disable on windows, bug in libuv
    if los.type() == 'win32' then return end

    function onExit() p('exit') end
    function onClose() p('close') end

    child = spawn('skfjsldkfjskdfjdsklfj')
    child:on('exit', expect(onExit))
    child:on('close', expect(onClose))
  end)
```

joerg-krause:
Yes, the `error` handler is called after `exec()` returns and after the `exit` handler.

The problem is that `exec()` returns a non-nil `child` and the `pid` of the child is an error message (e.g. `ENOMEM: not enough memory`), whereas I would expect that the child return by `exec()` is `nil`.

squeek502:
Yeah, it's a bit of a weird pattern. I prefer the `nil, err` return style, but it looks like this is setup for the `:on('error', function() end)` pattern exclusively, and changing that to `nil, err` would be a breaking change (and at the very least would need the tests updated). Perhaps as an alternative, the `Process` returned could handle the error case a bit better? For example, [this line](https://github.com/luvit/luvit/blob/455d2f1a41a511f46667180a2570d65bc436f839/deps/childprocess.lua#L151) could instead do something like `if not handle then em:setError(pid) else em:setPid(pid) end`?

By the way, it's very easy to get `uv.spawn()` to return `nil, err` by just giving it an invalid command to run; it will return `ENOENT: no such file or directory`.

joerg-krause:
I agree, I prefer the `nil, err` return style. On the hander hand, `child:on(...)` would need to be guarded by checking `if child then child:on(...) end`.

The real issue here is the lack of documentation. I often need to study the sources to understand how functions should be used.

So, closing this issue as there is nothing really wrong about the error handling, but just a misunderstanding of things actually work.

PR #1063, opened 2019-04-02
Bump lit/luvi versions to latest for testing

squeek502:
This is only for CI purposes. I'll rebase this branch as we fix things, so our goal is to make the CI pass on this branch.

See #1058 & #1060

PR #1067, opened 2019-04-03
test: Improve test-process.lua

zhaozg:
Now, test-process.lua not pass on windows, because uv_spawn fail with spawn process, and return `ENOMEM: not enough memory`.

I found https://github.com/nodejs/node/issues/25484 and https://support.microsoft.com/en-gb/help/947246/you-may-receive-an-out-of-memory-error-message-because-of-the-desktop.

Maybe we need to wait upstream to fix that problems.

@squeek502 
Ref https://github.com/luvit/luvit/pull/1063 https://github.com/luvit/luvit/issues/1058


squeek502:
Updating the tests here means that there is a breaking change in `childprocess`, right? This seems like the wrong approach. We should be trying to fix `childprocess` so that the tests pass, not update the tests to adhere to how `childprocess` happens to work now.

In other words, if I have a Luvit script that uses `childprocess.spawn`, then updating Luvi would break that script. Instead, we should try to make it so the script still works the same with the new Luvi.

zhaozg:
The child process mechanism has not actually changed, just code optimization. In addition, the previous test cases are not perfect enough to handle error events.

Subprocesses are more complex in libuv or luvit. In the future, the implementation of luvit can be simplified as appropriate. If multiple objects and handles are released and automatically closed, it is necessary to rethink some optimal solutions.

I am not getting enough free time at the moment, and can't put enough energy into it. If I introduce a mistake, I am very sorry.

squeek502:
No worries, thanks for making these fixes.

I'll do some more investigation/testing to see if what I was saying is true.

PR #1070, opened 2019-05-18
update http module

zhaozg:
* only close socket when ServerResponse not need to keepAlive
* use exampels/http-server.lua to show a keepalive benchmark
* remove examples/spped-http.lua


deps/http.lua
```diff
@@ -187,8 +187,10 @@ function ServerResponse:finish(chunk)
   last = last .. (self.encode("") or "")
   local function maybeClose()
     self:emit('finish')
-    self.socket:_end()
-    collectgarbage()
+    if not self.keepAlive then
+      self.socket:_end()
+      collectgarbage()
```
rphillips: a collectgarbage here should probably be removed to allow the user to manage garbage collection if necessary.

PR #1078, opened 2019-06-29
fix unclosed handle in childprocess module

zhaozg:
When a spawned child process exit, stdout and stderr will be closed as expected, but stdin keep open, and not fire close event. This because stdout and stderr care read event, close when EOF, but stdin as write pipe, that don't get error when not data to output, so need to close stdin explicitly. 
This will make all CI pass.

### DIFF
--- a/deps/childprocess.lua
+++ b/deps/childprocess.lua
@@ -126,6 +126,7 @@ local function spawn(command, args, options)
     em.exitCode = code
     em.signal = signal
     em:emit('exit', code, signal)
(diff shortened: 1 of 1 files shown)


zhaozg:
This is a sub job for https://github.com/luvit/luvit/issues/1058.

zhaozg:
https://github.com/luvit/luvit/pull/944

zhaozg:
follow https://github.com/luvit/luvit/commit/f499ade75cf3d8d09658e4f293c8a959f68065f4

PR #1079, opened 2019-07-01
update ssl to support session reuse for TLSV1.3

zhaozg:
Ths is just a intermediate state. The final solution need wait lua-openssl/luvi upgrade.

https://wiki.openssl.org/index.php/TLS1.3#Sessions says 

>  In TLSv1.3 sessions are not established until after the main handshake has completed. The server sends a separate post-handshake message to the client containing the session details. Typically this will happen soon after the handshake has completed, but it could be sometime later (or not at all).

> The old SSL_get1_session() and similar APIs may not operate as expected for client applications written for TLSv1.2 and below. Specifically if a client application calls SSL_get1_session() before the server message containing session details has been received then an SSL_SESSION object will still be returned, but any attempt to resume with it will not succeed and a full handshake will occur instead. In the case where multiple sessions have been sent by the server then only the last session will be returned by SSL_get1_session(). Calling SSL_get1_session() after a 2 way shutdown will give a resumable session if any was sent. You can check that a session is resumable with SSL_SESSION_is_resumable().

> Client application developers should consider using the SSL_CTX_sess_set_new_cb() API instead. This provides a callback mechanism which gets invoked every time a new session is established. This can get invoked multiple times for a single connection if a server sends multiple session messages.

> Note that SSL_CTX_sess_set_new_cb() was also available in previous versions of OpenSSL. Applications that already used that API will still work, but they may find that the callback is invoked at unexpected times, i.e. post-handshake.

> An OpenSSL server will immediately attempt to send session details to a client after the main handshake has completed. The number of tickets can be set using SSL_CTX_set_num_tickets. To server applications this post-handshake stage will appear to be part of the main handshake, so calls to SSL_get1_session() should continue to work as before.

> If a client sends it's data and directly sends the close notify request and closes the connection, the server will still try to send tickets if configured to do so. Since the connection is already closed by the client, this might result in a write error and receiving the SIGPIPE signal. The write error will be ignored if it's a session ticket. But server applications can still get SIGPIPE they didn't get before.

### DIFF
--- a/deps/tls/common.lua
+++ b/deps/tls/common.lua
@@ -109,7 +109,8 @@ end
 -------------------------------------------------------------------------------
 
 local Credential = Object:extend()
-function Credential:initialize(secureProtocol, defaultCiphers, flags, rejectUnauthorized, context)
+function Credential:initialize(secureProtocol,defaultCiphers, flags,
+                               rejectUnauthorized, context, isServer)
   self.rejectUnauthorized = rejectUnauthorized
   if context then
     self.context = context
@@ -161,7 +162,7 @@ function TLSSocket:initialize(socket, options)
 
   self.options = options
   self.ctx = options.secureContext
-  self.server = options.isServer
+  self.server = options.server
   self.requestCert = options.requestCert
   self.rejectUnauthorized = options.rejectUnauthorized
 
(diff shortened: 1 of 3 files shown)


zhaozg:
Ref https://github.com/luvit/luvit/issues/1058

PR #1085, opened 2019-10-03
http: ignore upgrade header when no upgrade listener is present

Timothy-:
If a server supports HTTP/2 it sends an "Upgrade: h2,h2c" header.
This should be ignored unless the client actually supports HTTP/2.

This is a breaking change for users that depend on the original behaviour.

### DIFF
--- a/deps/http.lua
+++ b/deps/http.lua
@@ -160,7 +160,6 @@ function ServerResponse:flushHeaders()
   head.code = statusCode
   local h = self.encode(head)
   self.socket:write(h)
-
 end
 
 function ServerResponse:write(chunk, callback)
@@ -346,12 +345,18 @@ function ClientRequest:initialize(options, callback)
   self.encode = codec.encoder()
   self.decode = codec.decoder()
 
+  if callback then
+    self:once('response', callback)
+  end
+
   local buffer = ''
   local res
 
   local function flush()
-    res:push()
-    res = nil
+    if res then
+      res:push()
+      res = nil
+    end
   end
 
   local socket = options.socket or net.createConnection(self.port, self.host)
@@ -363,11 +368,6 @@ function ClientRequest:initialize(options, callback)
     self.connected = true
     self:emit('socket', socket)
 
-    local function onEnd()
-      -- Just in case the stream ended and we still had an open response,
-      -- end it.
-      if res then flush() end
-    end
     local function onData(chunk)
       -- Run the chunk through the decoder by concatenating and looping
       buffer = buffer .. chunk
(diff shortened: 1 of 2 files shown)


SinisterRectus:
Is there any way you can make this not breaking? Maybe an initialization option for automatic upgrading that is true by default?

SinisterRectus:
Although maybe that's not a great idea if this is correcting behavior. Is this upgrade behavior documented somewhere?

Timothy-:
I've changed the implementation to be more in line with nodejs:
https://nodejs.org/dist/latest-v12.x/docs/api/http.html

The normal `http.request(opts, cb)` should still work and will ignore an upgrade header in the response. You can also listen on the response event: `req:on('response', cb)`

The upgrade and connect are special cases since the http codec must be detached. In these cases the 'response' event will not be emitted.

If you want to implement websockets, you need to listen for an upgrade request: 
`req:on('upgrade', cb)`
If you want to use http CONNECT, you need to listen for the connect event:
`req:on('connect', cb)`

deps/http.lua
```diff
@@ -363,51 +369,37 @@ function ClientRequest:initialize(options, callback)
...
-                break
               end
+                return self:emit(evt, res, socket, event)
+              elseif self.method == 'CONNECT' or res.statusCode == 101 then
+                return self:destroy()
             end
-            if self.method == 'CONNECT' then
-              self:emit('connect', res, socket, event)
             end
+            self:emit('response', res)
```
SinisterRectus: There are some indentation issues here, can you fix that?

deps/http.lua
```diff
@@ -363,51 +369,37 @@ function ClientRequest:initialize(options, callback)
...
-              -- If the request upgrades the protocol then detatch the listeners so http codec is no longer used
-              local is_upgraded
-              if res.headers.upgrade then
-                is_upgraded = true
-                socket:removeListener("data", onData)
-                socket:removeListener("end", onEnd)
+              end
+            if self.method == 'CONNECT' or res.headers.upgrade then
+              local evt = self.method == 'CONNECT' and 'connect' or 'upgrade'
+              if self:listenerCount(evt) > 0 then
```
SinisterRectus: Could you explain what this count check is for?

Timothy-: See [nodejs http client source](https://github.com/nodejs/node/blob/v12.11.1/lib/_http_client.js#L479).

If you are listening on "upgrade" it means that you expect an upgrade event. If you are not listening on that event, the upgrade header should be ignored. If you are not listening on "upgrade" and the server somehow sends a 101 "switching protocols" then the connection will be destroyed.

Nodejs doc:

> Emitted each time a server responds to a request with an upgrade. If this event is not being listened for and the response status code is 101 Switching Protocols, clients receiving an upgrade header will have their connections closed.

Timothy-:
AppVeyor finally passed.
It seems for windows setting `host = '127.0.0.1'` in a `http.request` is required, but not for linux.

SinisterRectus:
Sorry for the lack of attention. I only use coro-http, so the behavior of this module is not something with which I am familiar. If someone else can confirm that this change is sufficiently implemented and non-breaking, we can pull it and maybe bump and release.

squeek502:
I think this is a worthwhile change. Just ran into it and the current functionality basically creates a silent failure on `Upgrade` headers being received (you get a response in the callback but then nothing else happens).

I think it's worth fixing the 'normal' functionality here (assuming most people are using `http` for http requests) in exchange for potentially breaking backwards compatibility with the abnormal cases (websockets via the `http` module, `CONNECT`).

PR #1086, opened 2019-10-03
Implements ustring versions of Lua 5.3's utf8.char and utf8.codepoint

SinisterRectus:
Here are two functions for converting between UTF8 characters and numerical codepoints for Luvit's ustring library.

- **ustring.char(...)** takes a list of codepoints and converts them to a new ustring, similarly to string.char or utf8.char
- **ustring.codepoint(ustr, i, j)** returns numbers for each indicated encoded character in a ustring, similarly to string.byte or utf8.codepoint

I've also cleaned up the formatting of the library, to make it more consistent with other Luvit libraries, and bumped the library to 2.1.0.

Comments and suggestions appreciated.
Happy Hacktoberfest


SinisterRectus:
I was poking around at the [luvi source](https://github.com/luvit/luvi/blob/master/src/luvi_compat.c) and I only now just realized that it includes Lua's utf8 library. Whoops. I suppose this PR isn't entirely necessary, but maybe it's still useful.

SinisterRectus:
Closing as redundant.

PR #1093, opened 2020-01-01
fix: remove extra argument to doWrite call

datatypevoid:
The current implementation passes in an empty string as the node implementation does for an encoding parameter. In luvit there is no encoding parameter, so this should not be passed.


SinisterRectus:
Interesting find. Do you have an example of some user code that would be fixed by this?

datatypevoid:
Not particularly, I was just reading through the source and noticed the extra parameter.

Issue #1092, opened 2020-01-01
Issues with code  stuck


giepher:
There's the test code
```
local Emitter = require('core').Emitter
local i= 0
local Client = Emitter:extend()
function Client:initialize()
  self:on('Method1', function ()
            print('Method1 -- '..i)
            i = i + 1
  end)
end
function Client:start()
while true do
  self:emit('Method1')
end -- END while
end
local x = Client:new()
x:start()
```

and output stucked like this:

```
Method1 -- 0
Method1 -- 1
Method1 -- 2
Method1 -- 3
...
Method1 -- 954
Method1 -- 955
Method1 -- 956
Metho
```
It's stuck the last line.  I test this because my server been stucked when I want print some log .

Is someone can  help?

SinisterRectus:
I'm not able to reproduce this. What platform and luvit version are you running?

giepher:
both mac and centos7 with  luvit 2.16,  At last it will display out of memory if I didn't kill the process.

giepher:
> I'm not able to reproduce this. What platform and luvit version are you running?

both mac and centos7 with luvit 2.16, At last it will display out of memory if I didn't kill the process.

squeek502:
You're getting out of memory errors with the exact code you posted?

giepher:
Method1 -- 118
Method1 -- 119
Method1 -- 120
Method1 -- 121
Method1 -- 122
Method1 -- not enough memory

giepher:
> You're getting out of memory errors with the exact code you posted?

Method1 -- 118
Method1 -- 119
Method1 -- 120
Method1 -- 121
Method1 -- 122
Method1 -- not enough memory
the "not enough memory"  shows after 1 or 2 minutes, the process memory increaced to 2G  my RAM is 16G.
it's ok when i comment the print

squeek502:
The version of LuaJIT that Luvit uses has a hard memory limit of 2GB so that part makes sense (see https://github.com/luvit/luvi/issues/211).

The lines not being printed might be something to do with your terminal and how it buffers output? Not sure. I can reproduce the out-of-memory error by running the code for quite a while, and it prints the whole time. Here's the number I get to:
```
Method1 -- 3449784
Method1 -- 3449785
not enough memory
```

Not sure why the memory is not being cleaned up, though. That might be a bug.

SinisterRectus:
A simplified example also increases memory consumption.
```lua
while true do
  collectgarbage()
  collectgarbage()
[...]

PR #1097, opened 2020-02-28
make console/tty write safe

zhaozg:
make p and print output full/safe.

### DIFF
--- a/deps/pretty-print.lua
+++ b/deps/pretty-print.lua
@@ -297,6 +297,26 @@ function dump(value, recurse, nocolor)
   return nocolor and strip(s) or s
 end
 
+local function console_write(fs, s)
+  s = s .. '\n'
+  if uv.guess_handle(uv.fileno(fs))=='tty' then
+    repeat
+      local n, e = uv.try_write(fs, s)
+      if n then
+        s = s:sub(n+1)
+        n = 0
+      else
+        if e:match('^EAGAIN') then
+          n = 0
+        else
+          assert(n, e)
+        end
+      end
+    until n==#s
+  else
+    uv.write(fs, s)
+  end
+end
 -- Print replacement that goes through libuv.  This is useful on windows
 -- to use libuv's code to translate ansi escape codes to windows API calls.
 function _G.print(...)
@@ -305,8 +325,7 @@ function _G.print(...)
   for i = 1, n do
     arguments[i] = tostring(arguments[i])
   end
-  uv.write(stdout, table.concat(arguments, "\t"))
-  uv.write(stdout, "\n")
+  console_write(stdout, table.concat(arguments, "\t"))
 end
 
 function prettyPrint(...)
@@ -315,8 +334,7 @@ function prettyPrint(...)
   for i = 1, n do
     arguments[i] = dump(arguments[i])
   end
-  uv.write(stdout, table.concat(arguments, "\t"))
-  uv.write(stdout, "\n")
+  console_write(stdout, table.concat(arguments, "\t"))
 end
 
 if uv.guess_handle(0) == 'tty' then


deps/pretty-print.lua
```diff
@@ -297,6 +297,16 @@ function dump(value, recurse, nocolor)
...
 end
 
+local function console_write(fs, s)
+  s = s .. '\n'
+  repeat
+    n = uv.try_write(fs, s)
+    if n then
+      s = s:sub(n+1)
+    end
+    n = 0
```
rphillips: next line does a comparison on n equaling the length of s, but n holds the write length. There needs to be a counter.

squeek502:
I think using `try_write` will make it possible for output to be truncated/dropped which is a worse problem than infinite loops causing memory exhaustion. Is there a different problem this is trying to solve? If so, could you provide a reproduction of that problem?

See the comments here: https://github.com/luvit/luvit/issues/1092#issuecomment-575507954

deps/pretty-print.lua
zhaozg: s = s:sub(n+1) will skip data has been write, all write will get s=''

zhaozg:
>I think using try_write will make it possible for output to be truncated/dropped which is a worse problem than infinite loops causing memory exhaustion. Is there a different problem this is trying to solve? If so, could you provide a reproduction of that problem?

I found uv_write will runcated/dropped data, when process:exit(), but try_write make all output. I almost forget about https://github.com/luvit/luvit/issues/1092#issuecomment-575507954.

squeek502:
Using `try_write` for `tty` makes some sense to me, but I feel like I don't have enough knowledge to know what the trade-offs are here.

@SinisterRectus? @rphillips? Thoughts?

SinisterRectus:
I looked at this in Node.js and I *think* they use try_write for tty.

SinisterRectus:
Looking at why we reroute `print`:
>Print replacement that goes through libuv.  This is useful on windows
>to use libuv's code to translate ansi escape codes to windows API calls.

Something I just discovered (in Luvi 2.16.0 on Windows) is that if you open stdio handles with luv, Lua's IO calls are affected. Interestingly, whether you open fd 2 or 3 or both, Lua's io.stdout and io.stderr are both affected. I don't understand enough about how this works under the hood to know what's going on here. Can we rely on this behavior instead of re-writing `print`?

```lua
local uv = require('uv')

local colors = {
  black   = 30,
  red     = 31,
  green   = 32,
  yellow  = 33,
  blue    = 34,
  magenta = 35,
  cyan    = 36,
  white   = 37,
}

local function colorize(str, color)
  return string.format('\27[%i;%im%s\27[0m', 1, color, str)
end

local function open(fd)
  local handle
  if uv.guess_handle(fd) == 'tty' then
    handle = assert(uv.new_tty(fd, true))
  else
    handle = uv.new_pipe(false)
    uv.pipe_open(handle, fd)
  end
  return handle
end

open(0)
open(1)
open(2)

print(colorize('print', colors.cyan))
io.stdout:write(colorize('io.stdout:write', colors.green), '\n')
io.stderr:write(colorize('io.stderr:write', colors.yellow), '\n')
error(colorize('error', colors.red))
```
image

squeek502:
@SinisterRectus that doesn't seem to be universal. Here's what I get on Windows 7:

wincolors

Do you get the colors if you just use print? I think at some point the Windows console added native support for certain ANSI escape sequences.

SinisterRectus:
No. Without initializing the tty handles, I get the same raw output on Windows 10 that you see there in Windows 7. Good to know that it is not universal.

zhaozg:
May I merge this? @squeek502 @rphillips @SinisterRectus

deps/pretty-print.lua
```diff
@@ -297,6 +297,26 @@ function dump(value, recurse, nocolor)
...
+    repeat
+      local n, e = uv.try_write(fs, s)
+      if n then
+        s = s:sub(n+1)
+        n = 0
+      else
+        if e:match('^EAGAIN') then
+          n = 0
+        else
+          assert(n, reason)
```
squeek502: Where is `reason` coming from? Should this be `e`?

squeek502:
I'm okay with trying this.

deps/pretty-print.lua
zhaozg: Yes

PR #1112, opened 2020-05-03
fix and simplify code for Emitter:listeners

zhaozg:


squeek502:
This doesn't look like the same functionality to me. Currently, the removal of listeners is done 'lazily', where listeners are set to `false` and only actually [removed (via `table.remove`) in `Emitter:emit`](https://github.com/luvit/luvit/blob/f8f4f87992f23efd75458014f4f3c8ea47c3cb3e/deps/core.lua#L250-L254).

I'm not sure why it's done that way, exactly, but I'm not sure I'm confident enough to make this change without a good reason.

deps/core.lua
```diff
@@ -238,45 +229,56 @@ function Emitter:emit(name, ...)
...
+      local h = handlers[i]
+      if type(callback) == "table" then
+        -- only remove once callback table
+        h = type(h) == 'table' and h == callback or false
+      elseif type(callback) == "function" then
+        -- remove once callback table or other callback function
+        h = type(h) == 'table' and h.callback == callback or h == callback
       end
       if h then
-        handlers_for_type[i] = false
```
zhaozg: Emitter:removeListener `handlers_for_type[i] = false` only affect the libuv next cycle.
Direct remove it is safe?

zhaozg: Should I not change this?

deps/core.lua
```diff
@@ -286,12 +288,7 @@ function Emitter:removeAllListeners(name)
   local handlers = rawget(self, "handlers")
   if not handlers then return end
   if name then
-    local handlers_for_type = rawget(handlers, name)
-    if handlers_for_type then
-      for i = #handlers_for_type, 1, -1 do
-          handlers_for_type[i] = false
-      end
-    end
+    rawset(handlers, name, {})
```
zhaozg: Same, next loop

deps/core.lua
```diff
@@ -301,15 +298,7 @@ end
 --  @param {String} name event name
 function Emitter:listeners(name)
   local handlers = rawget(self, "handlers")
-  if not handlers then
-    return 0
```
zhaozg, line 305: Should return empty table

zhaozg:
Done

PR #1113, opened 2020-05-03
simplify threadpool

zhaozg:


squeek502:
I don't see the point of this change, and it's backwards incompatible since it's possible that people rely on the returned `worker` extending from `Object`.

zhaozg:
I originally wanted to increase efficiency, but Negligible

PR #1114, opened 2020-05-03
fix test-process, make sure signal handled

zhaozg:
uv loop may stop before handle signal, so a timer will make all signal handle as expect.

### DIFF
--- a/tests/test-process.lua
+++ b/tests/test-process.lua
@@ -16,6 +16,8 @@ require('tap')(function(test)
   end)
 
   test('signal usr1,usr2,hup', function(expect)
+    -- Note: Luvit unrefs the global process' signals, so we need to make sure to keep the loop alive long enough
+    -- for the signals to actually be handled. This is what the timer is doing in this test.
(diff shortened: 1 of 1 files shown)


squeek502:
I'd like to reproduce https://github.com/luvit/luvit/issues/1099 locally and try to understand it a bit more before reviewing this.

zhaozg:
I can reproduce #1099  on my macOS when with https://github.com/luvit/luvit/pull/1115

tests/test-process.lua
```diff
@@ -16,17 +16,30 @@ require('tap')(function(test)
   end)
 
   test('signal usr1,usr2,hup', function(expect)
-    local onHUP, onUSR1, onUSR2
+    local onHUP, onUSR1, onUSR2, iCount
```
squeek502, line 19:
```suggestion
    -- Note: Luvit unrefs the global process' signals, so we need to make sure to keep the loop alive long enough
    -- for the signals to actually be handled. This is what the timer is doing in this test.
    local onHUP, onUSR1, onUSR2, iCount
```

PR #1115, opened 2020-05-03
make console/tty write safe

zhaozg:
reactive, but pending https://github.com/luvit/luvit/pull/1114


zhaozg:
after mege #1114, this PR ci must be restart manually.

PR #1116, opened 2020-05-03
Add Socket:buffer_size api

zhaozg:


deps/net.lua
```diff
@@ -154,6 +154,16 @@ function Socket:keepalive(enable, delay)
   uv.tcp_keepalive(self._handle, enable, delay)
 end
 
+function Socket:buffer_size(send_size, recv_size)
```
squeek502, line 157: I think this should be called `Socket:set_buffer_size` and then a `Socket:get_buffer_size` should be added too:

```lua
function Socket:get_buffer_size()
  local send_size = uv.send_buffer_size(self._handle)
  local recv_size = uv.recv_buffer_size(self.handle)
  return send_size, recv_size
end
```

SinisterRectus: Agree, but it should be camelCase.

SinisterRectus:
Moving this to a top-level comment because I find inline comments awkward.

I can't find a corresponding API for this in node, so I guess we're on our own.

I think it may be worth making this API more verbose:

```lua
Socket:getSendBufferSize()
Socket:getRecvBufferSize()
Socket:setSendBufferSize(size)
Socket:setRecvBufferSize(size)
```

You'd also want to do some input validation on `size`, since its type and value can change the behavior of the internal call.

```lua
assert(type(size) == "number" and size > 0, "Size must be a number greater than 0")
```

zhaozg:
Done

PR #1117, opened 2020-05-04
Signal removeListener fixes

squeek502:
Something I noticed while looking into #1099

> Before this, calling removeListener would unconditionally close the signal, even if there were other listeners still listening for that signal type. Now, the signal will only be closed once there are no listeners left.

---

This will conflict with https://github.com/luvit/luvit/pull/1114 and https://github.com/luvit/luvit/pull/1112. I'll fix the conflicts once those are merged.

### DIFF
--- a/deps/core.lua
+++ b/deps/core.lua
@@ -256,7 +256,9 @@ function Emitter:emit(name, ...)
 end
 
 -- Remove a listener so that it no longer catches events.
+-- Returns the number of listeners removed, or nil if none were removed
 function Emitter:removeListener(name, callback)
+  local num_removed = 0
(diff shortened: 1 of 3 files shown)


deps/core.lua
```diff
@@ -271,13 +273,20 @@ function Emitter:removeListener(name, callback)
...
     for i = #handlers_for_type, 1, -1 do
       handlers_for_type[i] = false
+      num_removed = num_removed + 1
     end
   end
+  if num_removed > 0 then
+    return num_removed
+  else
+    return nil
+  end
```
zhaozg, line 289: I like `return num_removed > 0 and num_removed or nil`

PR #1125, opened 2020-11-02
ustring.lua: empty string ~= empty list

shawwn:
On Lua 5.3.5, `ustring.new("") == {}`

This is rather unexpected behavior, so this PR fixes it by ensuring both types are the same.

(I'm open to suggestions on how better to fix this.)

### DIFF
--- a/deps/ustring.lua
+++ b/deps/ustring.lua
@@ -225,20 +225,8 @@ end
 _meta.__index = ustring
 
 function _meta.__eq(ustr1,ustr2)
-    local len1 = #ustr1
-    local len2 = #ustr2
-    if len1 ~= len2 then return false end
-
-    for i = 1,len1 do
-        if ustr1[i] ~= ustr2[i] then return false end
-    end
-
-    if len1 == 0 and len2 == 0 then
-        if getmetatable(ustr1) ~= getmetatable(ustr2) then
-            return false
-        end
-    end
-    return true
+    if #ustr1 ~= #ustr2 then return false end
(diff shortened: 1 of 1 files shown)


SinisterRectus:
Lua 5.3 omits the type and metatable check in `___eq`. What about the non-empty string case? For example, is `ustring.new("a") == {"a"}`?

shawwn:
You're right! `ustring.new("a") == {"a"}` evaluates to `true`.

I've made another attempt. Can you think of any clever ways of breaking this?

SinisterRectus:
I guess that works.

SinisterRectus (approved):

SinisterRectus:
I'm going to merge this as a compatibility fix for Lua 5.3/5.4. It should have no effect on the behavior in LuaJIT (default) or Lua 5.1 or Lua 5.2.

Lua 5.1:
> "eq": the == operation. The function getcomphandler defines how Lua chooses a metamethod for comparison operators. A metamethod only is selected when both objects being compared have the same type and the same metamethod for the selected operation.

Lua 5.2:
> "eq": the == operation. The function getequalhandler defines how Lua chooses a metamethod for equality. A metamethod is selected only when both values being compared have the same type and the same metamethod for the selected operation, and the values are either tables or full userdata. 

Lua 5.3 / Lua 5.4:
> __eq: the equal (==) operation. Behavior similar to the addition operation, except that Lua will try a metamethod only when the values being compared are either both tables or both full userdata and they are not primitively equal. The result of the call is always converted to a boolean.

PR #1126, opened 2020-11-27
Localize p in stream_readable.lua.

TSnake41:
Fix an issue in stream_readable.lua where roundUpToNextPowerOf2() overrides the global variable p (which is also in fact, less efficient than using local variable).

Not sure if it does impact any project there.


squeek502 (approved):
Good catch, thanks!

SinisterRectus (approved):

PR #1127, opened 2020-11-29
Redesign http-codec.lua state machine

TSnake41:
Use integers instead of functions as decoder state machine mode.

It may improve JIT efficiency on this code as JIT compiler may be able to use a static jump table instead of having to deal with function variables (pointers) and may be able to inline some decoder parts into the built decoder.

In a simple program that does only contain empty functions comparing the two state machine designs, it appears to very significatively improve performance. But in the case of http-codec, it doesn't appear to have a visible impact on decoder performance.

### DIFF
--- a/deps/http-codec.lua
+++ b/deps/http-codec.lua
@@ -162,12 +162,14 @@ end
 local function decoder()
 
   -- This decoder is somewhat stateful with 5 different parsing states.
-  local decodeHead, decodeEmpty, decodeRaw, decodeChunked, decodeCounted
+  local decodeHeadMode, decodeEmptyMode, decodeRawMode,
+        decodeChunkedMode, decodeCountedMode = 1, 2, 3, 4, 5
+
   local mode -- state variable that points to various decoders
   local bytesLeft -- For counted decoder
 
   -- This state is for decoding the status line and headers.
-  function decodeHead(chunk)
+  local function decodeHead(chunk)
     if not chunk then return end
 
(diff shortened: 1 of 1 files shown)


SinisterRectus:
How about keeping an array of the functions and indexing them with `mode`?

rphillips:
Functions are usually pointers... It would be nice to have some benchmarks of the various scenarios to figure out the benefits.

SinisterRectus:
> Functions are usually pointers... It would be nice to have some benchmarks of the various scenarios to figure out the benefits.

@TSnake41 Hi, I know this is an old PR, but would you be interested in writing any perf tests? If not, I or someone else can take a stab at it.

TSnake41:
> > Functions are usually pointers... It would be nice to have some benchmarks of the various scenarios to figure out the benefits.
> 
> @TSnake41 Hi, I know this is an old PR, but would you be interested in writing any perf tests? If not, I or someone else can take a stab at it.

I haven't worked with luvit for a long time, so it will be better that someone else takes this task.

SinisterRectus:
It looks like you've only refactored the decoder. There is also an encoder to consider.

There is also another [http-codec in the lit project](https://github.com/luvit/lit/blob/master/deps/http-codec.lua) that is a bit different from the one here. While that doesn't need the same changes, it would make the code even more divergent if it is not refactored.

I am all for optimizations, and I like your idea, but I think this one is going to be too much of a burden for little gain. Maybe we can consider it in the future if it's ever written for "luvit 3.0", especially if we move away from a closure API.

PR #1138, opened 2021-05-20
Remove Freenode and add Discord link

rphillips:
Change the official channels to be Discord or the mailing list.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -14,7 +14,7 @@ See the main project webpage for more details. <https://luvit.io/>
 
 ## Need Help?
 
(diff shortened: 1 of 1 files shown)


SinisterRectus (approved):
I was never really invested in IRC, so I don't have an opinion on the freenode situation. I've also never read our mailing list. It looks like the most recent posts are old and spam.

I have no plans to stop managing the Discord server. We have some nice people there and it's occasionally active, although still somewhat quiet. If you're reading this and you haven't joined, feel free to pop in and say hi.

PR #1139, opened 2021-06-03
Wrap main in a coroutine

truemedian:
This is technically a breaking change, so I'd like to see whether this has to wait until Luvit 3 (if it ever exists) or if it'll make it into Luvit 2.x.

However, this change allows for running libraries like coro-http without having to wrap an entire file in a coroutine.

### Changes

* The second return of `coroutine.running` is now `false` in any mainland code, the main thread is (mostly) inaccessible.
* `coroutine.yield` will now work successfully instead of throwing a C-call boundary error in mainland code.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -45,8 +45,8 @@ return function (main, ...)
 
   local args = {...}
   local success, err = xpcall(function ()
-    -- Call the main app
-    main(unpack(args))
+    -- Call the main app inside a coroutine
+    coroutine.wrap(main)(unpack(args))
 
     -- Start the event loop
     uv.run()


Bilal2453:
I've definitely wanted this all the time, in fact, I am quite surprised that Luvit 2 didn't wrap the main chunk in a coroutine already.

Though I guess this is technically a breaking change too, and if Luvit 3 was ever going to be a thing, I would love a bit more into this, such as explicitly exposing the main thread, and handling edge cases such as stack-overflow in case of errors (since I would guess `coroutine.wrap` would leave a similar effect to the last fixed bug where error-stack would be eaten by coro-*), and similar stuff.

creationix:
I like this!  I'm just wondering what exactly the breaking change is.

truemedian:
It would break a theoretical program that expects the main function to be run in the main thread, so the most accessible change would be that the second return of `coroutine.running` is now false everywhere that isn't a libuv callback.

creationix:
I kinda want to just do this and see what happens.  We'll want to test it manually on the most popular luvit projects and make sure they don't break, but I don't think they will.

truemedian:
I absolutely agree that it should break absolutely no code, because it shouldn't be relying upon such behavior in the first place. However, I just wanted to at least get some feedback from people who might've seen code that did it for whatever reason.

Bilal2453:
Oh lol, let's go for it then! I would love this change not gonna lie.

creationix (approved):

PR #1154, opened 2021-08-06
 support Luajit or PUC Lua v5.x

zhaozg:
1. remove table.getn
2. support load Lua version > 5.1, remove setfenv
3. replace table.foreach with pairs
4. unpack or table.unpack


zhaozg:
Maybe we need build a luvi with lua 5.4 to test.

PR #1142, opened 2021-06-25
Release Prep for Luvit v2.18.0

creationix:
- luvi: bump to v2.12.0 [Tim Caswell]
- lit: update to 3.8.5 [Tim Caswell]
- luvit: add the changelog [Tim Caswell]
- Wrap main in a coroutine [@truemedian]

PR #1155, opened 2021-08-29
Merge pull request #1142, HTTP CODEC

leXinshou:
Recently, I have been facing a server breaking error which is reason 
```
http-codec.lua:115: attempt to concatenate local 'reason' (a nil value)
```
Which triggers only when there is no reason provided or STATUS_CODES doesn't exist. As there a few documented STATUS CODES but are not present in the http_codec, sending "Unknown Reason" would be better than breaking the whole server process because of one single reason " reason is nil ".

It shouldn't affect the core, however it will prevent breaking the server.

### DIFF
--- a/deps/http-codec.lua
+++ b/deps/http-codec.lua
@@ -111,7 +111,7 @@ local function encoder()
       assert(path and #path > 0, "expected non-empty path")
       head = { item.method .. ' ' .. item.path .. ' HTTP/' .. version .. '\r\n' }
     else
-      local reason = item.reason or STATUS_CODES[item.code]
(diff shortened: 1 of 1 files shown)


squeek502:
Here's what the status line spec has to say for context: https://datatracker.ietf.org/doc/html/rfc7230#section-3.1.2


> The reason-phrase element exists for the sole purpose of providing a
> textual description associated with the numeric status code, mostly
> out of deference to earlier Internet application protocols that were
> more frequently used with interactive text clients.  A client SHOULD
> ignore the reason-phrase content.

So this solution seems okay in that respect.

squeek502 (approved):

PR #1156, opened 2021-09-06
Repl: rows implementation and fixes

### LINKED ISSUE #1107: Luvit's repl weird behavior when writing a long code
Hitting backspace/arrow keys while writing a code that's longer than the terminal width will result in duplicated lines. 

According to @truemedian on luvit's discord server, this is caused by this [https://github.com/luvit/luvit/blob/master/deps/readline.lua#L99-L106](code), as it resets it to the left edge but never goes up.

-----------------------------------

***Pic of the issue:***
-----------------------------------
ChLjvpHn4f

-----------------------------------
**luvit version:** *2.17.0*
-----------------------------------
**luvi version:** *v2.10.1*
-----------------------------------

Bilal2453:
Fixes #1107

This PR tries to not change how readline worked in the past, while introducing tty rows to the system, this is achieved by refreshing the whole line with all of its row (by making sure to always start writing from the left of the very first row). Unlike the way NodeJS's refreshLine clear the screen, this will only clear said line, which means we must handle each position change differently and separately; that might be a might unclean, but it is the best way available to us without having to rewrite the whole thing. I will walk you through each of those changes step by step.

1. [Editor.getCursorOffset](https://github.com/Bilal2453/luvit/commit/28d03e27fc445badb587f2d49132203bc06f078a#diff-1e96b4ccbce65e70ba904c2c3a2748b0db6d77ec8552820997ebc0d0b17660f5R87-R93): This is the way I am using to measure where exactly the cursor is, we could've used `ESC[6n` instead, or used the NodeJS of manually counting each character width, but both will require a good portion of code while the first requiring a bit of rewrite of how onKey works. This seems to be accurate as long as UTF-8 unicodes are not presented (more about that later). On the other hand, there is no need for NodeJS's way of doing it since Luvit repl does not allow tabs characters in a row entered by the user. Although do note that this makes use of `Editor.columns` which is NOT updated on resize events, meaning all you need to break this is resizing the terminal; I do plan on implementing this later on.

2. [refreshLine's command](https://github.com/Bilal2453/luvit/commit/28d03e27fc445badb587f2d49132203bc06f078a#diff-1e96b4ccbce65e70ba904c2c3a2748b0db6d77ec8552820997ebc0d0b17660f5R109-R115): Couple required changes were made here, first of them forcing a new line to fix # , making sure the cursor goes to a new line when needed. Second of them is using `ESC[0J` which is not required but it makes more sense, plus it simplifies handling the deletion of first character of a row.

3. [refreshLine's row refreshing](https://github.com/Bilal2453/luvit/commit/28d03e27fc445badb587f2d49132203bc06f078a#diff-1e96b4ccbce65e70ba904c2c3a2748b0db6d77ec8552820997ebc0d0b17660f5R118-R127): This will basically force the cursor to go up first initial line, to start the rewrite from there. As you can see because of the way we are handling this we will have to handle each case separately, having to handle a row up (moveLeft) and a row down (moveRight).

4. [Editor.insertAbove](https://github.com/Bilal2453/luvit/commit/28d03e27fc445badb587f2d49132203bc06f078a#diff-1e96b4ccbce65e70ba904c2c3a2748b0db6d77ec8552820997ebc0d0b17660f5R131-R140): Since the user might trigger this when they are at first row while having more rows, we need to handle this a bit differently than used to. Neither this nor previous implementation handled more than a single row insertion.

5. [Editor.insert](https://github.com/Bilal2453/luvit/commit/28d03e27fc445badb587f2d49132203bc06f078a#diff-1e96b4ccbce65e70ba904c2c3a2748b0db6d77ec8552820997ebc0d0b17660f5R153-R158): Making sure the cursor is always visible, trigger a refreshLine when a new row is required.

### DIFF
--- a/deps/readline.lua
+++ b/deps/readline.lua
@@ -84,6 +84,13 @@ function History.new()
 end
 
 local Editor = {}
+function Editor:getCursorOffset()
+  local position = self.position
+  local columns = self.columns
+  local rows = math.floor((position + 1) / columns)
+  local cols = position - (columns * math.floor(position / columns))
+  return rows, cols
+end
 function Editor:refreshLine()
   local line = self.line
   if self.cover then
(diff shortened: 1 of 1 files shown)


Bilal2453:
Lastly, I need  to mention when this fix will not work.

1. If a Unicode is used in the line, this seems to be true for both this fix and previous Luvit implementation. Not sure why but it seems like Luvit messes the count on it or something. It does not cause big problems although.

2. When terminal is resized, this totally breaks everything. We must handle onResize and update `Editor.columns`.

3. When the line is spanned upon multiple pages, that's, no enough space to show all rows at once. The biggest deal seems to happen when insert-between is triggered. NodeJS suffer from the same bug.

4. In some rare cases `Editor.getHistory` seems to eat a row, not sure why, was not able to replicate it a second time.

PR #1176, opened 2022-05-02
thread: fix thread.equals using wrong name

Bilal2453:
Luv exposes the libuv method under the name [uv.thread_equal](https://github.com/luvit/luv/blob/master/src/thread.c#L400).
While `thread.equals` tries to index a non-existence method.


Bilal2453:
huh somehow a commit from my other PR got in here

PR #1180, opened 2022-08-15
add tostring() on pegmatch

Sezei:
Avoids a few unexpected errors.

### DIFF
--- a/deps/json.lua
+++ b/deps/json.lua
@@ -673,7 +673,7 @@ function json.use_lpeg ()
     local npos
     local t, nt = {}, 0
     repeat
-      obj, cont, npos = pegmatch (ArrayContent, str, pos, nullval, state)
+      obj, cont, npos = pegmatch (ArrayContent, tostring(str), pos, nullval, state)
       if not npos then break end
       pos = npos
       nt = nt + 1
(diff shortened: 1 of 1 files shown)


Bilal2453:
Uhhh, I don't know if we should be patching dkjson like this... Worst case scenario is adding couple of type checking

Duckwhale:
It would be good to have a minimale example (or test case) to see what problem this change actually fixes :)

Sezei:
> It would be good to have a minimale example (or test case) to see what problem this change actually fixes :)

A prime example for it would be my discord bot I made not too long ago, [Blueberry](https://github.com/Sezei/Blueberry); For some reason it always responded with "bad argument #2 to 'pegmatch'; expected string, got boolean (or sometimes nil)"; happened with most commands that returned an embed. The commands themselves worked, but just returned the pegmatch error. Modifying my json dep to have tostring(str) solved it instantly.

truemedian:
That error only happens when you're not passing a json string to `json.decode` in the first place. Using `tostring` here will more often than not result in a parsing error anyway because the only thing that will correctly `tostring()` into a valid json payload is `true` and `false`

Sezei:
> That error only happens when you're not passing a json string to `json.decode` in the first place. Using `tostring` here will more often than not result in a parsing error anyway because the only thing that will correctly `tostring()` into a valid json payload is `true` and `false`

As you would probably see in the source code, I do not use any booleans in any if my responses in the first place.

PR #1181, opened 2022-09-23
add string.split

RemRBLX:
this makes luvit work like roblox


squeek502:
Would need a more compelling reason to add this. You are welcome to define `string.split` in your own code, there's no reason it needs to be part of Luvit.

PR #1182, opened 2022-10-10
fix url parse

zhaozg:
fix auth field order

### DIFF
--- a/deps/url.lua
+++ b/deps/url.lua
@@ -17,7 +17,7 @@ limitations under the License.
 --]]
 --[[lit-meta
   name = "luvit/url"
-  version = "2.1.2"
+  version = "2.1.3"
   dependencies = {
     "luvit/querystring@2.0.0",
   }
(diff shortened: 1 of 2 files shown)


squeek502:
Could you add test cases too?

deps/url.lua
```diff
@@ -126,6 +126,7 @@ local function parse(url, parseQueryString)
     search = search,
     query = query,
     auth = auth,
+    href = href,
     hash = hash
   }
   parsed.href = URL.format(parsed)
```
squeek502, line 132: Does this not overwrite the `href` set above?

zhaozg: Yes, will fixit.

zhaozg:
test update

tests/test-url.lua
```diff
@@ -86,4 +86,24 @@ require('tap')(function(test)
...
+    local parsed = url.parse('http://user:passwd@host:81/')
+    assert(parsed.pathname == '/')
+    assert(parsed.auth == 'user:passwd')
+    assert(parsed.path == '/')
+    assert(parsed.host == 'host:81')
+    assert(parsed.href == 'http://user:passwd@host:81/')
+    assert(parsed.port == '81')
+    assert(parsed.protocol == 'http')
+    assert(parsed.hostname == 'host')
+  end)
```
squeek502, line 108: These can be added to the `parseTests` table instead:

This can be added to the `parseTests` table above instead:

```
  ["/dir/@file"] = {href = '/dir/@file', path = '/dir/@file', pathname = '/dir/@file'},
  ["http://user:passwd@host:81/"] = {href = 'http://user:passwd@host:81/', protocol = 'http', auth = 'user:passwd', host = 'host:81', hostname = 'host', port = 81, path = '/', pathname = '/'},
```

squeek502 (approved):

PR #1183, opened 2022-10-21
Run completion in repl in coroutine

creationix:
I was testing an object that has metamethods for `__pairs` that get triggered when I use autocomplete in the repl.  But my data fetcher needs to be run in a coroutine context.  This PR fixes that.  We have coroutine everywhere else, but missed it the key handler.


squeek502 (approved):
CI failure looks unrelated / intermittent (since one `ubuntu-latest` run succeeded and one failed; might be worth creating an issue for it though).

PR #1184, opened 2022-10-23
Removes value `nil`

UrNightmaree:
Removes value `nil` after `startRepl`, `combo` and `script` because it's useless.

### DIFF
--- a/deps/helpful.lua
+++ b/deps/helpful.lua
@@ -72,7 +72,7 @@ return function (prefix, mod)
     __index = function (table, wanted)
       if type(wanted) ~= "string" then return end
       local closest = math.huge
-      local name = nil
+      local name
       for key in pairs(table) do
         local distance = string.levenshtein(key, wanted)
         if distance < closest then
--- a/deps/path/base.lua
+++ b/deps/path/base.lua
@@ -169,7 +169,7 @@ end
 function Path:resolve(...)
   local paths = {...}
   local resolvedpath = ""
-  local resolveddrive = nil
+  local resolveddrive
   local isabsolute = false
   for i=#paths, 1, -1 do
     local path = paths[i]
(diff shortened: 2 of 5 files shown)


ravener:
Maybe you could've done the other ones too, here's all I could find:
```
~/l/luvit  find . -type f -exec grep -i --color=always 'local .* = nil' \{\} +
./deps/helpful.lua:      local name = nil
./deps/path/base.lua:  local resolveddrive = nil
./deps/readline.lua:  local consumedKeys = nil
./deps/stream/stream_readable.lua:  local er = nil
./main.lua:  local startRepl = nil
./main.lua:  local combo = nil
./main.lua:  local script = nil
```

RiskoZoSlovenska:
What's wrong with initializing variables to `nil` when declaring them? IMO it makes it clear that a different value may be set later on but that the default value should be `nil` otherwise, whereas a bare declaration implies that the variable isn't initialized and thus isn't being used yet (usually due to scoping, such as in [pathjoin.lua](https://github.com/luvit/luvit/blob/master/deps/pathjoin.lua#L10))

DocRevive:
> a different value may be set later on but that the default value should be `nil` otherwise

> a bare declaration implies that the variable isn't initialized and thus isn't being used yet

These mean the same thing. We know the declaration implicitly initializes the variable as `nil`. The difference you note is an obscure and contrived matter of interpretation that, if it was a real cause of uncertainty, could be fixed by a comment. Some people consistently prefer one over the other for certain reasons (to one of which I say verbosity is not readability) but the link you provide shows this is in fact inconsistent. There's no concrete reason for this difference.

SinisterRectus:
I consider this mostly a stylistic opinion. Looking through the code, it seems like the declaration `local var` is more common than the explicit definition `local var = nil` so I'm okay accepting this for consistency.

PR #1187, opened 2023-02-04
Replace code deprecated in Lua 5.1 and 5.2

SinisterRectus:
Towards making Luvit less dependent on Lua versioning, I've replaced some code that is deprecated in earlier versions of Lua.

### Lua 5.1

> Function `table.setn` was deprecated. Function `table.getn` corresponds to the new length operator (`#`); use the operator instead of the function.

I replaced `table.getn` calls with `#`. Luvit doesn't use `table.setn`.

> Functions `table.foreach` and `table.foreachi` are deprecated. You can use a for loop with `pairs` or `ipairs` instead. 

I replaced `table.foreach` with `pairs` loops. Luvit doesn't use `table.foreachi`.

Both `unpack` and `table.unpack` exist in LuaJIT with `DLUAJIT_ENABLE_LUA52COMPAT`, but `table.unpack` does not exist in stock LuaJIT and `unpack` does not exist in Lua 5.3+. I converted `unpack` calls to indexings where possible or added conditional `unpack` definitions elsewhere.

> Function `loadstring` is deprecated. Use `load` instead; it now accepts string arguments and are exactly equivalent to `loadstring`.

Both `loadstring` and `load` exist in stock LuaJIT and have the same 5.2 behavior; `loadstring` does not exist in 5.3+. I replaced all `loadstring` calls with `load` calls.

> Functions `setfenv` and `getfenv` were removed, because of the changes in environments.

All calls to `setfenv` are on loaded functions, so I just moved the environment tables into the `load` calls. All calls to `getfenv` are used to define `require` in thread callbacks, which have their own global environment, so I changed those to simple assignments and it seems to work the same.

### DIFF
--- a/deps/childprocess.lua
+++ b/deps/childprocess.lua
@@ -201,9 +201,9 @@ local function _exec(file, args, options, callback)
     maxBuffer = 4 * 1024,
     signal = 'SIGTERM'
   }
-  table.foreach(opts, function(k, v)
+  for k, v in pairs(opts) do
     if not options[k] then options[k] = v end
-  end)
+  end
 
   local child = spawn(file, args, options)
 
--- a/deps/codec.lua
+++ b/deps/codec.lua
@@ -29,6 +29,7 @@ limitations under the License.
 
 local uv = require('uv')
 local assertResume = require('utils').assertResume
+local unpack = unpack or table.unpack ---@diagnostic disable-line: deprecated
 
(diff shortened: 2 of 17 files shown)


SinisterRectus:
See luvit/luvi#272 for a related change.

zhaozg (approved):

PR #1188, opened 2023-02-20
buffer: make sure memory is zeroed out

Bilal2453:
As I just noticed while reviewing my [luvit-meta](https://github.com/Bilal2453/luvit-meta) definitions of the Buffer class, doing the following:

```lua
local buf = buffer.Buffer:new(16)
print(buf:inspect())
```
will print a string with random contents, I first thought `malloc` did zero-out the memory it allocates but apparently not.

Thanks for @truemedian to point that out + telling me to use `calloc` instead of `memset` as this is apparently faster than doing another call to `memset`, this should "fix" it.  "Fix it" assuming it wasn't an intended behavior.

### DIFF
--- a/deps/buffer.lua
+++ b/deps/buffer.lua
@@ -34,7 +34,7 @@ local Object = core.Object
 local instanceof = core.instanceof
 
 ffi.cdef[[
-  void *malloc (size_t __size);
+  void *calloc (size_t nmemb, size_t __size);
   void free (void *__ptr);
 ]]
 
@@ -49,11 +49,11 @@ local C = ffi.os == "Windows" and ffi.load("msvcrt") or ffi.C
 function Buffer:initialize(length)
   if type(length) == "number" then
     self.length = length
-    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.malloc(length)), C.free)
+    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(1, length)), C.free)
   elseif type(length) == "string" then
     local string = length
     self.length = #string
-    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.malloc(self.length)), C.free)
+    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(1, self.length)), C.free)
     ffi.copy(self.ctype, string, self.length)
   else
     error("Input must be a string or number")
--- a/deps/buffer.lua
+++ b/deps/buffer.lua
@@ -49,11 +49,11 @@ local C = ffi.os == "Windows" and ffi.load("msvcrt") or ffi.C
 function Buffer:initialize(length)
   if type(length) == "number" then
     self.length = length
-    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(1, length)), C.free)
+    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(length, 1)), C.free)
   elseif type(length) == "string" then
     local string = length
     self.length = #string
-    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(1, self.length)), C.free)
+    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(self.length, 1)), C.free)
     ffi.copy(self.ctype, string, self.length)
   else
     error("Input must be a string or number")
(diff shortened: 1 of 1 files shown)


deps/buffer.lua
```diff
@@ -49,11 +49,11 @@ local C = ffi.os == "Windows" and ffi.load("msvcrt") or ffi.C
 function Buffer:initialize(length)
   if type(length) == "number" then
     self.length = length
-    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.malloc(length)), C.free)
+    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(1, length)), C.free)
```
truemedian, line 52: Just a correctness thing, the size should be `1` and the length should be `length`. You're allocating `length` 1-byte items. Not 1 `length`-byte items

Bilal2453: Hm, not sure if that analogy makes more sense to me than the other, mind explaining more on why is it "Allocating `length` 1-byte items` instead of it being a single `length` item?

truemedian: Because it's an array of bytes (`size=1`), not a single `size=len` item.

Bilal2453:
The Lit server failed at the wrong time, the build-failure is unrelated and is caused by an http error on Lit server side 😬

deps/buffer.lua
```diff
@@ -49,11 +49,11 @@ local C = ffi.os == "Windows" and ffi.load("msvcrt") or ffi.C
 function Buffer:initialize(length)
   if type(length) == "number" then
     self.length = length
-    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.malloc(length)), C.free)
+    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(length, 1)), C.free)
   elseif type(length) == "string" then
     local string = length
     self.length = #string
-    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.malloc(self.length)), C.free)
+    self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(self.length, 1)), C.free)
```
squeek502, line 56: This can still use `malloc`, since the full contents will be overwritten by the contents of the string anyway.

squeek502:
Seems somewhat reasonable but I'm not totally sure what the use-case/benefit of zero-initialized memory here is.

truemedian:
Zeroed memory is required if you want to have any semblance of reasonable initialized data. `malloc` is allowed to return memory that has been freed but not yet reclaimed by the OS, which can and will contain random data. If you read any part of the buffer that hasn't been written to (like the end), you're stepping into completely garbage data.

squeek502:
I guess it depends on the use-case. In my mind reading those bytes at all would likely be indicative of a bug, so the data being zeros or garbage wouldn't make much of a difference, but if the buffer is zero-initialized then I can see how it'd open up opportunities for depending on that behavior.

I'm good with merging this once https://github.com/luvit/luvit/pull/1188#discussion_r1112439914 is addressed.

SinisterRectus:
FWIW, Node has different constructors `alloc` and `allocUnsafe` for the different behaviors.

deps/buffer.lua
Bilal2453: oh oops missed that

Bilal2453:
One possible solution is to leave the current behavior, well document it, and tell people to instead use `Buffer:new(('\0'):rep(size))` if they want a zeroed out memory.  Not exactly fan of that option although

Following what Sinister said, we could also have an additional argument in the constructor, `zeroMem`, on false (default) just use `malloc`, otherwise zero it out for those who do depend on it being zeros.

SinisterRectus:
Something like that might be okay. Note that direct construction of Buffers without using the static constructors was deprecated in Node 6. <https://nodejs.org/docs/latest-v6.x/api/buffer.html#class-buffer>

deps/buffer.lua
```diff
@@ -35,6 +35,7 @@ local instanceof = core.instanceof
 
 ffi.cdef[[
   void *malloc (size_t __size);
+  void *calloc (size_t nmemb, size_t __size);
```
Bilal2453, line 38: Why do the arguments here start with `__`?  Should the nmemb argument be `__nmemb` too?

rphillips:
Based on https://nodejs.org/api/buffer.html#static-method-bufferallocsize-fill-encoding, the buffer should likely be prefilled with zeros, and perhaps extend the API to be `Buffer.initialize(size[, fill])`, where fill defaults to `0` (zero).

squeek502 (approved):

PR #1208, opened 2024-06-17
remove or indicate all mandatory luajit dependencies from luvit

truemedian:
This PR removes the luajit ffi and jit libraries as mandatory dependencies for all but 2 luvit libraries: dns and buffer, both of which use ffi as part of their operation.

The general migration pattern was: anything that used `jit.os` or `ffi.os` now uses the luvit `los` library.

The `los` library now tries to use a few different methods to find the OS, starting with `jit.os` and `ffi.os`. Then it tries to use `uname -s`. Then it tries the `OS` environment variable.

This means `los` will now work in a non-luajit environment, this is not a breaking change.

Migration of `require` was handled differently, because it does not actually depend on any other luvit libraries. To this end: how luvit require chooses a temporary directory has changed, it is now OS-agnostic. The binary extension is now inferred from the first entry of the `package.cpath`, or the path separator present in `package.config` if a cpath entry is not present or does not have a file extension.

Accompanying these changes:

- The version of `luvit/require` has been bumped to `2.2.4`
  - This version no longer has any dependency on luajit
- The version of `luvit/los` has been bumped to `2.0.1``
  - This version no longer has any dependency on luajit
- The version of `luvit/dns` has been bumped to `2.0.5``
  - This version only has a dependency on luajit for Windows
- The version of `luvit/buffer` has been bumped to `2.1.3`
  - This version has a mandatory dependency on luajit, but it now has an informative error.

### DIFF
--- a/deps/los.lua
+++ b/deps/los.lua
@@ -18,7 +18,7 @@ limitations under the License.
 
 --[[lit-meta
   name = "luvit/los"
-  version = "2.0.0"
+  version = "2.0.1"
   license = "Apache 2"
   homepage = "https://github.com/luvit/luvit/blob/master/deps/los.lua"
   description = "Tiny helper to get os name in luvit."
--- a/deps/los.lua
+++ b/deps/los.lua
@@ -25,19 +25,70 @@ limitations under the License.
   tags = {"os"}
 ]]
 
-local jit = require('jit')
+local uv = require('uv')
 
-local map = {
-  ['Windows'] = 'win32',
-  ['Linux'] = 'linux',
-  ['OSX'] = 'darwin',
-  ['BSD'] = 'bsd',
-  ['POSIX'] = 'posix',
-  ['Other'] = 'other'
+local function get_os_name()
+  -- shortcuts for luajit if available
+  local has_jit, jit = pcall(require, 'jit')
+  if has_jit and jit and jit.os then
+    return jit.os
+  end
+
+  local has_ffi, ffi = pcall(require, 'ffi')
+  if has_ffi and ffi and ffi.os then
+    return ffi.os
+  end
+
+  -- Handles windows
+  if os and os.getenv then
+    if os.getenv('OS') then
+      return os.getenv('OS')
+    end
+  end
+
+  -- use libuv provided uname if possible, but it's not always available
+  if uv.os_uname then
+    local info = uv.os_uname().sysname
+    if info then
+      return info.sysname
+    end
+  end
+
+  -- Handles most other posix platforms
+  if io and io.popen then
+    local uname_child = io.popen('uname -s')
+    if uname_child then
+      local os_name = uname_child:read('*l')
+      uname_child:close()
+      return os_name
+    end
+  end
+
+  return 'other'
+end
+
+local os_patterns = {
+  ['windows'] = 'win32',
+  ['linux'] = 'linux',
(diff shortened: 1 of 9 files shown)


deps/pathjoin.lua
```diff
@@ -9,12 +9,7 @@
 
 local getPrefix, splitPath, joinParts
 
-local isWindows
-if _G.jit then
-  isWindows = _G.jit.os == "Windows"
-else
-  isWindows = not not package.path:match("\\")
-end
+local isWindows = require("los").type() == "win32"
```
SinisterRectus, line 12: This is part of the bigger issue of code replicates mentioned in #1166, but can you also apply this patch to luvipath.lua in luvi, or is it not worth the effort?

deps/los.lua
```diff
@@ -18,26 +18,67 @@ limitations under the License.
...
-  version = "2.0.0"
+  version = "2.0.1"
   license = "Apache 2"
   homepage = "https://github.com/luvit/luvit/blob/master/deps/los.lua"
   description = "Tiny helper to get os name in luvit."
   tags = {"os"}
 ]]
 
-local jit = require('jit')
+local function get_os_name()
```
SinisterRectus, line 28: Can we use `uv_os_uname`? It was added in https://github.com/luvit/luv/pull/313.

Bilal2453: Seems possible? but that should be used alongside the jit requirement, since `uv_os_name` was introduced in libuv 1.25 and older versions don't have it.

truemedian: I went ahead and added a check or uv_os_uname, but as Bilal mentioned it's Libuv 1.25+, so I've kept a call to the uname utility as a fallback in case everything else fails.

deps/pathjoin.lua
truemedian: A similar patch could be applied to luvi, but the isWindows check is one of the things that likely can't change; `los` isn't available for luvi because its a luvit library.

deps/require.lua
```diff
@@ -29,14 +29,14 @@ local luvi = require('luvi')
...
 
 local realRequire = _G.require
 
-local tmpBase = os == "Windows" and (env.get("TMP") or uv.cwd()) or
-                                    (env.get("TMPDIR") or '/tmp')
-local binExt = os == "Windows" and ".dll" or ".so"
+local tmpBase = env.get("TMPDIR") or env.get("TMP") or env.get("TEMP") or (uv.fs_access("/tmp", "r") and "/tmp") or uv.cwd()
+
+local first_cpath = package.cpath:match("[^" .. package.config:sub(3, 3) .. "]+")
+local binExt = (first_cpath and first_cpath:match("%.[^.]+$")) or (package.config:sub(1, 1) == "\\" and ".dll" or ".so")
```
SinisterRectus, line 39: What is the reason for this `binExt` change?

truemedian: Require has no dependencies on luvit libraries, so it can't use `los` to get the os name. The `binExt` change makes `require` take the system binary extension from the first entry in package.cpath; and only use a naive check if the system uses backslashes as a backup.

This primarily makes it trivial to support macos, which may use exclusively `.dylib` libraries.

SinisterRectus (approved):
Looks good to me.

squeek502 (approved):

PR #1219, opened 2024-11-23
buffer: add Lua tables implementation

Bilal2453:
Remove the Buffer class dependency on FFI as part of the efforts of making Luvit luajit-agnostic.

This change will simply use a table for `buffer.ctype` instead of an FFI malloc-ated ctype object and implements the necessary changes (mostly in __newindex and in toString) to account for differences.

Other changes also include:
- added a check for the invalid initialization `Buffer:new(-1)`, as negative lengths mean nothing.
- added a check in `toString` for when `i > j`, (e.x. `buffer:toString(3, 1)`), using FFI this would error, but with the table implementation we had to do this check ourselves. 

Also added some test cases, mostly related to when things should be raising errors.


deps/buffer.lua
```diff
@@ -27,42 +27,74 @@ limitations under the License.
...
+      self.ctype = ffi.gc(ffi.cast("unsigned char*", C.malloc(self.length)), C.free)
+      ffi.copy(self.ctype, content, self.length)
+    else
+      self.ctype = ffi.gc(ffi.cast("unsigned char*", C.calloc(self.length, 1)), C.free)
+    end
+  else
+    local raw = {}
+    if content then
+      for i = 1, self.length do
+        raw[i - 1] = castChar(content:sub(i, i):byte())
```
truemedian, line 89: string.byte is guaranteed to return a number between 0 and 255

deps/buffer.lua
```diff
@@ -207,9 +243,18 @@ Buffer.writeInt32LE = Buffer.writeUInt32LE
...
+  i = i and i - 1 or 0
+  j = j or self.length
+  if (i < 0 or i > self.length) or (j and j > self.length) or (i > j) then
+    error("Range out of bounds")
+  end
+  if not hasFfi then
+    if self.length <= 0 then
+      return ''
+    end
+    return char(unpack(self.ctype, i, j - 1))
```
truemedian, line 255: using unpack here is unsound, in Luajit and Lua 5.1 it will fail for any buffer of size > MAXCSTACK (by default 8000)

deps/buffer.lua
```diff
@@ -27,42 +27,74 @@ limitations under the License.
...
 
---avoid bugs when link with static run times lib, eg. /MT link flags
-local C = ffi.os == "Windows" and ffi.load("msvcrt") or ffi.C
+local function castChar(val)
+  if val >= 0 and val <= 0xff then
+    return val
+  else
+    return band(val, 0xff)
+  end
+end
```
truemedian, line 64: This function is unnecessary, `band(val, 0xff)` exhibits the desired behavior

truemedian (changes requested):

deps/buffer.lua
Bilal2453: The unpack usage here is meant as an optimization, this way we only have to call `string.char` once.

deps/buffer.lua
Bilal2453: Correct, the if statement is meant as an early return to not unnecessarily AND the value. I will remove it though, seems to have a negative effect on performance rather than a good one

deps/buffer.lua
Bilal2453: Here are the benchmarks for `table.unpack` vs `table.concat`:
```
Buffer library benchmarks using Lua tables (JIT ON, table.unpack)
Creating 10e5 16-byte buffers: 0.291
Creating 10e5 128-byte buffers: 0.767
Creating 10e5 16-byte buffers with content: 0.326
Creating 10e5 128-byte buffers with content: 0.838
Creating and calling toString on 10e5 16-byte buffers: 0.442
Creating and calling toString on 10e5 128-byte buffers: 1.256
Write operations with 10e6 iteration: 0.0030000000000001
Read operations with 10e8 iteration: 0.241
```
```
Buffer library benchmarks using Lua tables (JIT ON, table.concat)
Creating 10e5 16-byte buffers: 0.362
Creating 10e5 128-byte buffers: 0.87
Creating 10e5 16-byte buffers with content: 0.31
Creating 10e5 128-byte buffers with content: 0.858
Creating and calling toString on 10e5 16-byte buffers: 0.818
Creating and calling toString on 10e5 128-byte buffers: 3.477
Write operations with 10e6 iteration: 0.0030000000000001
Read operations with 10e8 iteration: 0.243
```
and the implementation is
```lua
  if not has_ffi then
    if self.length <= 0 then
      return ''
    end
    local tbl = {}
    for c = i, j - 1 do
      tbl[#tbl+1] = char(self.ctype[c])
    end
    return table.concat(tbl)
  end
```

Seems like using unpack is worth the performance gain here, what do you think?

truemedian: I don't doubt that passing unpack directly into string.char is fast. I said it is unsound, it *will* fail for buffer sizes >8KB and may fail earlier for strange configurations (anything that reduces MAXCSTACK).

So either this needs to be an optimization with a fallback, or we drop the performance gain here entirely in favor of having code that works.

truemedian (approved):

Issue #1141, opened 2021-06-25
travis-ci.org is dead


squeek502:
> > [Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.](https://travis-ci.org/)
> 
> We'll probably want to migrate our travis stuff to Github Actions rather than travis-ci.com

Forgot to post an issue for Luvit, too (see also https://github.com/luvit/luv/issues/550, https://github.com/luvit/luvi/issues/249)

PR #1225, opened 2024-12-06
add a github action for coveralls

### LINKED ISSUE #1220: Coveralls wrong coverage percentage?
When inspecting [the coverage](https://coveralls.io/github/luvit/luvit) it appears to me Coveralls is counting the test files as part of the code that needs to be tested. Coveralls Lua support is provided by [luacov-coveralls](https://github.com/moteus/luacov-coveralls), looking at their [example](https://github.com/moteus/lua-travis-example) we can see that Coveralls [is reporting](https://coveralls.io/github/moteus/lua-travis-example) their code without including files from `test` directory unlike our case.

I believe we are currently missing a `.luacov` configuration file in `tests`? Their example seems to include one which has [include statements](https://github.com/moteus/lua-travis-example/blob/master/test/.luacov#L23-L30). We should probably test the proper way of reporting coverage with [luacov](https://github.com/lunarmodules/luacov) locally first then committing the right configurations.

Bilal2453:
This brings back the luacov-coveralls support as a github action, which was removed with the Travis deprecation in https://github.com/luvit/luvit/pull/1146, further contributing toward https://github.com/luvit/luvit/issues/1141.

Also fixes #1220.

### Code notes

Usually we run the coverage with `make cover`, which is `luvi . -- -l luacov tests/run.lua`, which means luacov needs to be ran from the root tree.  Luacov searches the current working directory for a `.luacov` config file, but we instead have that in `tests/.luacov` (for a cleaner tree), the action will set `tests/.luacov` as the config file with `luacov-coveralls -c tests/.luacov` which seems to be enough.

### Setup notes

This action is triggered on new commits as well as pull requests, it is also possible to manually trigger it with a dispatch.

Note that the Github token is used instead of a Coveralls token, which means you do not have to setup anything, the Github token should be already provided to Actions through `secrets.GITHUB_TOKEN`.  The only thing that might need setup is after merging the PR the root tree prefix in Coveralls might need to be updated to `/home/runnder/work/luvit/luvit/`.

### DIFF
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -1,0 +1,36 @@
+name: "Coverage analysis"
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  luacov:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Luvit
+        run: make luvit
+      - name: Setup Lua & Luarocks
+        env:
+          LUA_VER: lua-5.1.5
+          LUAROCKS_VER: luarocks-3.11.1
+        run: |
+          curl https://www.lua.org/ftp/$LUA_VER.tar.gz | tar xz
+          cd $LUA_VER
+          make linux && make install INSTALL_TOP=../install
+          cd ..
+          curl -L https://luarocks.github.io/luarocks/releases/$LUAROCKS_VER.tar.gz | tar xz
+          cd $LUAROCKS_VER
+          ./configure --with-lua=$GITHUB_WORKSPACE/$LUA_VER/install --prefix="./install"
+          make && make install
+          ln ./install/bin/luarocks $GITHUB_WORKSPACE/luarocks
+      - name: Install luacov & luacov-coveralls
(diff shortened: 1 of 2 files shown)


Bilal2453:
For a successful test run see https://coveralls.io/builds/71234240, not sure if the current PR failure is related to this not being merged yet.

Bilal2453:
@squeek502 can you give this a look

squeek502:
Tried re-running the CI since it seemed like the failure was a flaky test, but now it's failing during "Setup Lua & Luarocks". Any idea what might be happening there?

Bilal2453:
Instead of building Luarocks from the tar file I am now installing it from apt, which should automatically satisfy the Lua and readline dependencies, we can't now control the version being installed but I think that is fine.

The build error seems to be unrelated, it is of a Windows test that has been failing on Linux for a while, here is a successful [test run](https://github.com/Bilal2453/luvit/actions/runs/12438079002/job/34729272221).

squeek502 (approved):

squeek502:
Thanks!

PR #1227, opened 2025-01-16
adopt creationix/pathjoin, bump resource for new pathjoin

truemedian:
This was missed during the adoption in the Lit repository but `pathjoin`'s home is here and it was never fixed here. `luvit/resource` was bumped to rely on the new `luvit/pathjoin` instead of `creationix/pathjoin`

### DIFF
--- a/deps/pathjoin.lua
+++ b/deps/pathjoin.lua
@@ -1,5 +1,5 @@
 --[[lit-meta
-  name = "creationix/pathjoin"
+  name = "luvit/pathjoin"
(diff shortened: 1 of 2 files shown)


squeek502 (approved):

SinisterRectus:
Why did this fail *after* merging? https://github.com/luvit/luvit/actions/runs/12819245136

Bilal2453:
This test can't fail due to this change. It likely failed due to network.

It failed after merge because it was re-triggered on `push`. Tests seem to be ran again when merged.

PR #1230, opened 2025-06-21
deps: update dkjson to 2.8

Bilal2453:
Update dkjson from 2.5 to 2.8.

Most notably in this support for Lua 5.3/5.4 has been added.

Surprisingly, this module hasn't had a sync with upstream in more than 10 years! (Although to be fair, it also hasn't received any updates from 2014 until 2021)

Why not move it into its own mirror like we are doing with lpeg/pcre/etc?

### DIFF
--- a/deps/json.lua
+++ b/deps/json.lua
@@ -1,7 +1,7 @@
 --[[lit-meta
   name = "luvit/json"
-  version = "2.5.2"
-  homepage = "http://dkolf.de/src/dkjson-lua.fsl"
+  version = "2.8"
+  homepage = "http://dkolf.de/dkjson-lua"
   description = "David Kolf's JSON library repackaged for lit."
   tags = {"json", "codec"}
   license = "MIT"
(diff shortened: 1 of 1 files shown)


zhaozg (dismissed):

deps/json.lua
```diff
@@ -714,17 +753,17 @@ function json.use_lpeg ()
       return obj, retpos
     end
   end
+  jsonlpeg.parse = json.decode
```
truemedian, line 756: This should be `jsonlpeg.decode` shouldn't it?

Bilal2453: Thanks for noticing that, would've went unnoticed forever 😅

squeek502:
To clarify, these changes seem to move `lpeg` from an optional dependency to a hard dependency?

That seems fine when used within `luvit`, but `luvit/json` installed separately via `lit` and then attempting to run with Lua/LuaJIT instead of `luvi` would cause a problem, right?

(not necessarily a blocker, just want to check if that's correct)

truemedian (approved):

Issue #1022, opened 2018-01-21
Relative requires in threads, also Windows paths


SinisterRectus:
I'm having trouble with relative requires from within a thread. Maybe I don't know enough about how luvit's require sets up the cwd. This is tested on Windows.

```lua
local thread = require('thread')

local moduleName = 'coro-http'

thread.start(function(dir, name)

	local function test(path)
		print(pcall(require, path))
	end

	local uv = require('uv')
	local pathJoin = require('pathjoin').pathJoin
	test(name) -- doesn't work
	test(pathJoin('./deps', name)) -- doesn't work
	test(pathJoin(dir, 'deps', name)) -- does not work
	test(pathJoin(uv.cwd(), 'deps', name)) -- works

end, module.dir, moduleName):join()
```

Is there a correct way to use relative requires in a thread?
Additionally, is reason why paths on Windows are [converted to Unix paths](https://github.com/luvit/luvit/blob/9a7708469f88b1d8a6c5ee7b5910ce78e0163aa4/deps/require.lua#L207-L208)?

squeek502:
> Additionally, is reason why paths on Windows are converted to Unix paths?

Seems like this is just normalization, so that require('C:\test.lua') and require('C:/test.lua') resolve to the same cached value.

creationix:
Sorry, don't remember the reason for this.

SinisterRectus:
I don't remember why I closed this. I'm not working on this at the moment, but the issue still exists, so I'll re-open.

SinisterRectus:
Same as #983, I think.

PR #1239, opened 2026-03-16
fix: detect Windows fixed paths

### LINKED ISSUE #1238: Fixed require does not work on Windows
## Description
This should work:
```lua
local mod = require("C:/a/b/c/module")  -- does not work
```

because on Unix, the equiv does work:
```lua
local mod = require("/home/a/b/c/module")  -- works
```

but the Windows one errors with:
```
module 'C:/a/b/c//module' not found:
        no field package.preload['C:/a/b/c//module']
        no file 'C:\Users\bilal\AppData\Local\Programs\LuaJIT\bin\lua\C:/a/b/c//module'
        no file 'C:\Users\bilal\AppData\Local\Programs\LuaJIT\bin\lua\C:/a/b/c//module'
        no file 'C:\Users\bilal\AppData\Local\Programs\LuaJIT\bin\C:/a/b/c//module'
        no file 'C:\Users\bilal\AppData\Local\Programs\LuaJIT\bin\C:/a/b/c//module'
...
```

## Explanation

The reason for this failure is `Module.resolve`, it tries to guess when to use `fixedRequire` by looking if the first character of the path is either `.` or `/`, in case of relative requires this would work because Windows uses the same prefix, but for absolute paths this would not work since Windows ones could start with any of `C:`/`D:` etc, so it falls back to a `moduleRequire`.

There are multiple other issues with Windows paths, but I believe this is the most major one.

Bilal2453:
Fixes #1238

Simply check if the second character is a `:`. This *should* be safe because module names shouldn't contain such a character unless it is a fixed path. Unix paths may contain it as part of the name, but we check for that first.

### DIFF
--- a/deps/require.lua
+++ b/deps/require.lua
@@ -17,7 +17,7 @@ limitations under the License.
 --]]
 --[[lit-meta
   name = "luvit/require"
-  version = "2.2.4"
+  version = "2.2.5"
   homepage = "https://github.com/luvit/luvit/blob/master/deps/require.lua"
   description = "Luvit's custom require system with relative requires and sane search paths."
   tags = {"luvit", "require"}
@@ -255,11 +255,16 @@ end
 function Module:resolve(name)
   assert(name, "Missing name to resolve")
   local debundled_name = name:match("^bundle:(.*)") or name
-  if debundled_name:byte(1) == 46 then -- Starts with "."
+  local prefix = debundled_name:byte(1)
+  -- If it starts with "." it's a relative path
+  if prefix == 46 then
     return fixedRequire(pathJoin(self.dir, name))
-  elseif debundled_name:byte(1) == 47 then -- Starts with "/"
+  end
+  -- If it starts with "/" or "x:/" it's a fixed path
+  if prefix == 47 or debundled_name:match('^%a:/') then
     return fixedRequire(name)
   end
+  -- Otherwise, it's a module name
   return moduleRequire(self.dir, name)
 end
 


Bilal2453:
(luckily enough, the important CI test failed due to Lit being reliable)

truemedian:
This is arguably more hacky than just checking for `/`. Use luvipath to check if getPrefix returns non-nil, which handles all forms of absolute paths on all platforms.

Bilal2453:
It is pretty hacky indeed.

I don't believe `luvi.path.getPrefix` will work here. Before we reach Module.resolve, the path is being (wrongfully) normalized to use `/` prefixes, even if it is a Windows path, take the following example of `getPrefix` outputs:
```
local getPrefix = luvi.path.getPrefix

getPrefix("C:\\a\\b\\c") --> C:\ (correct)
getPrefix("/a/b/c") --> / (correct)

getPrefix("C:/a/b/c") --> nil (wrong)
```
Is there something else we could depend on instead of this? perhaps fix the normalization first?

squeek502:
Something like this?

```lua
  elseif debundled_name:byte(1) == 47    -- Starts with "/"
    or debundled_name:byte(1) == 92      -- or "\"
    or debundled_name:match("^%a:[\\/]") -- or "x:\" or "x:/"
  then
```

Bilal2453:
Perhaps yeah.
```lua
    or debundled_name:byte(1) == 92      -- or "\"
```
what case would this be handling? Unix paths should always be `/`.

Bilal2453:
Updated it to use match. This works for my use case, I might've not made clear what I meant with the path normalization thing.

I was trying to do `require(pathjoin(module.dir, 'path/to/native/library'))`. But `module.dir` is specifically being normalized on Windows to not contain `\`.

The reason for that seems to be the fact you can't actually require anything at all using `\` separator, i.e. using a path like this `require("C:\\a\\b\\c")` does not work (even when it is properly detecting fixed paths), but with `/` it does work. So it is pretty important to make sure anything you pass to require on Windows is not using the `\\` path separator or it won't work.

truemedian (dismissed):

squeek502:
> what case would this be handling? Unix paths should always be `/`.

It'd be a "rooted" or "root relative" path on Windows, which is relative to the root of the current working directory. So `\foo` or `/foo` would be resolved to `C:\foo` if the CWD is on `C:\`, or `D:\foo` if the CWD is on `D:\`, etc. Very rarely used intentionally but techinically one of the [many path types](https://chrisdenton.github.io/omnipath/Overview.html) supported on Windows.

> The reason for that seems to be the fact you can't actually require anything at all using \ separator, i.e. using a path like this require("C:\\a\\b\\c") does not work

Should this patch then just look for `/`, e.g. ` debundled_name:match('^%a:/')`?

Bilal2453:
> It'd be a "rooted" or "root relative" path on Windows

I see, I will assume libuv is capable of handling those without us having to add more code for it.

> Should this patch then just look for /, e.g.  debundled_name:match('^%a:/')?

For now yes, I am not entirely sure if this is a bug or intended behaviour, if it is intended then yeah we should only check for `/`. The only other mention of this I found is in https://github.com/luvit/luvit/issues/1022, creationix says he doesn't remember why it's like this, which kinda imply it was intended? I will remove it for now so we don't introduce a new "base of expectations", and if we later discover this is unintended we can fix it and add it back.
From a quick look on the rest of code, it *seems* to me like it's explicitly assuming UNIX paths for some reason, which is totally not the behaviour one would expect, I once accidentality set `module.path/dir` to a path with `\` separators (I was normalizing the path), `require` completely broke on Windows, what a mess 😅

Should we make a super-issue to track all require-related bugs? They are currently splattered, and I believe some of the things I mentioned here aren't even reported.

Bilal2453:
- Bumped the patch version.
- Made things look slightly nicer (maybe?) by removing duplicate string.byte calls.
- Removed the `\` check from `x:/`. (needs investigation)

squeek502:
Sorry to keep bouncing around on this, but if the `match` call doesn't include `\`, then the byte check should not include it either imo. I'd say just go with the most common stuff for now (absolute POSIX and drive-absolute Windows, POSIX path separators only) and leave the rest for later if it's ever needed ([Windows paths are kind of a nightmare](https://github.com/ziglang/zig/pull/25993))

```lua
  -- If it starts with "/" or "x:/" it's a fixed path
  if prefix == 47 or debundled_name:match('^%a:/') then
```

Bilal2453:
No worries, this is a good point! I should've probably discussed this in an issue first.

squeek502 (approved):

